PR #5, opened 2016-05-06
WorkInProgress :: Feature/steelpenny log dir migration

cdaringe:
# problem statement
- logs are wild
- logs are confusing
- logs arent in the new coins homebase with guaranteed write access

# solution
- update logging deps
- map logs to `/var/log/steelpenny/`


cdaringe:
Thanks for the boostie @rssk.  I'll take your follow up commit tidy as review.  If that's not ok, please reopen!

PR #8, opened 2016-05-26
Minor read me update

drewlandis:


cdaringe:
@drewlandis, i'd say we should be free to merge doc updates at will.  what do you think?

drewlandis:
Sounds good to me.  Thanks!

PR #24, opened 2016-10-18
Dataexchange request get statistics

drewlandis:
#### asana
https://app.asana.com/0/8755179914444/197010238354300

# problem statement
Create `getStatistics` controller for dcwebservices rewrite

# Note
make sure you run the following liquibase at https://github.com/MRN-Code/db_schema/pull/234 OR just run this on your DB
`GRANT EXECUTE ON FUNCTION dxdba.dx_get_statistics_f(integer) TO node_api;`


packages/steelpenny/src/test/integration/dataexchange.js
```diff
@@ -54,8 +54,49 @@ describe('Dataexchange tests', () => {
       });
     });
 
+    describe('GET getStatistics', () => {
+      it('Gets statistics for empty request', () => { // eslint-disable-line arrow-body-style
+        return apiClient.dataexchange.get(
+          'request',
+          'getStatistics',
+          '{"_topGroup":{"_parentId":null,"_filters":[],"_children":[],"_type":"and"},"_label":"Untitled Request","_modalities":{"MR":{"id":1,"label":"MR","count":0,"dx_filter_order":1},"Assessments":{"id":112,"label":"Assessments","count":0,"dx_filter_order":2},"Studies":{"id":143,"label":"Studies","count":0,"dx_filter_order":3},"Subjects":{"id":142,"label":"Subjects","count":0,"dx_filter_order":4}},"_readOnly":false}'
```
ruwang: Not sure it matters. Could we reduce the string length but constructing an Object before `return` statement, then using `Object.toString` for the param?

packages/steelpenny/src/test/integration/dataexchange.js
```diff
@@ -54,8 +54,49 @@ describe('Dataexchange tests', () => {
...
+        .then((response) => {
+          const result = response.data;
+          result.data[0].should.deep.equal({
+            result:
+            {
+              MR: { id: 1, label: 'MR', count: 76, dx_filter_order: 1 },
+              Assessments: { id: 112, label: 'Assessments', count: 0, dx_filter_order: 2 },
+              Studies: { id: 143, label: 'Studies', count: 1, dx_filter_order: 3 },
+              Subjects: { id: 142, label: 'Subjects', count: 38, dx_filter_order: 4 },
+            },
```
mstone121: It's quite possible that these numbers will change over time.

drewlandis: I grabbed on old query that was already submitted, so the query itself cannot change.  You are correct that these numbers _can_ but unless ABIDE adds new data or removes data, then they should be relatively stable.  I'm open to changing the test and/or using a different query, but at least this sets up a good test bed from which to run tests against `getStatistics`

packages/steelpenny/src/test/integration/dataexchange.js
drewlandis: Good suggestion!  I updated to use `JSON.stringify(obj)` and the tests continue to pass 👍

packages/steelpenny/src/lib/controllers/dataexchange.js
```diff
@@ -13,17 +13,17 @@ module.exports.get = {
...
+  description: 'API for DataExchange front end',
   auth: false,
   validate: {
     options: {
       allowUnknown: true,
     },
     query: {
       objType: joi.string().allow(['request']),
-      action: joi.string().allow(['loadFilterableModalities']),
+      action: joi.string().allow(['loadFilterableModalities', 'getStatistics']),
```
swashcap: 👍

packages/steelpenny/src/lib/controllers/dataexchange/request.js
```diff
@@ -18,6 +19,40 @@ const loadFilterableModalities = (bookshelf) => {
     }, {}));
 };
 
-module.exports = {
-  loadFilterableModalities,
+internals.getStatistics = (bookshelf, query) => {
+  const requestJson = JSON.parse(query.json);
```
swashcap: Doesn’t hapi parse this for you? If not, we should move this up to the root data exchange controller such that all sub-controllers receive a parsed JSON object.

packages/steelpenny/src/lib/controllers/dataexchange/request.js
```diff
@@ -18,6 +19,40 @@ const loadFilterableModalities = (bookshelf) => {
...
+        0::bigint AS mr,
+        0::bigint AS assessments,
+        0::bigint AS studies,
+        0::bigint AS subjects
+    `;
+  }
+
+  return Promise.all([
+    bookshelf.knex.raw(qryStr),
+    internals.loadFilterableModalities(bookshelf),
```
swashcap: It seems silly to inject Bookshelf into a method. Is there a better way to resolve plugins in hapi land, @rssk?

packages/steelpenny/src/lib/controllers/dataexchange/request.js
```diff
@@ -18,6 +19,40 @@ const loadFilterableModalities = (bookshelf) => {
...
+  ])
+  .then(response => {
+    const resp = response[0].rows[0];
+    const lfm = response[1];
```
swashcap: Confusing variable name. To clarify, I’d change this to `filterableModalities` or the like.

packages/steelpenny/src/lib/controllers/dataexchange.js
```diff
@@ -13,23 +13,31 @@ module.exports.get = {
...
       json: joi.string(),
     },
   },
   handler: (clientRequest, reply) => {
+    const query = {};
+    Object.keys(clientRequest.query).forEach((key) => {
```
swashcap: Another way to think about this: you’re transforming `clientRequest.query`’s `json` property only. Why create a new data structure that iterates through all of `clientRequest.query`’s properties?

I’d write this like so:

```js
const query = 'json' in clientRequest.query ?
  Object.assign({}, clientRequest.query, {
    json: JSON.parse(clientRequest.query.json)
  }) :
  clientRequest.query;
```

Using `Object#assign` ensures you’re not mutating arguments, a big no-no.

**However!** Because Hapi _probably_ isn’t wrapping your handler in a `try...catch` you should do it explicitly. Even if there’s a generic `try...catch`, adding a custom one allows you to include a better error message. So:


```js
let query;

if ('json' in clientRequest.query) {
  try {
    query = Object.assign({}, clientRequest.query, {
      json: JSON.parse(clientRequest.query.json),
    });
  } catch (error) {
    return boom.badData('Request JSON could\'t be parsed', error);
  }
} else {
  query = clientRequest.query;
}
```

(Don’t forget to `require('boom')` up top.)

packages/steelpenny/src/test/integration/dataexchange.js
```diff
@@ -19,6 +19,21 @@ function setApiClient(client) {
 // Set should property of all objects for BDD assertions
 chai.should();
 
+function getQueryJson() {
```
swashcap: You’re always calling `JSON.stringify` on this function’s return value. Why not roll it into the function itself?

packages/steelpenny/src/lib/controllers/dataexchange.js
swashcap: Also, to be _extra_ safe, you might check that `clientRequest.query.json` is never an empty string. Those might be meaningful requests otherwise.

packages/steelpenny/src/test/integration/dataexchange.js
drewlandis: I like doing it this way that way we can operate on the object as an object, rather than as a string.

PR #25, opened 2016-10-18
Dataexchange filter getformattributes

ruwang:
#### asana
https://app.asana.com/0/8755179914462/197010238354303

# problem statement
`getFormAttributes` controller for dcwebservice rewrite.


packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,70 @@
...
+        _attributes[attributeId]._options.push(
+          {
+            optionId: parseInt(response.rows[i].option_id, 10),
+            optionLabel: response.rows[i].option_label,
+            count: parseInt(response.rows[i].count, 10),
+          }
+        );
+        _attributes[attributeId]._parentId = parseInt(requestJson._parent_id, 10);
+      }
+      for (const key in _attributes) {
```
jwtlake: Might want to use `of` instead of `in`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,70 @@
...
+        if (_attributes.hasOwnProperty(key)) {
+          ret.push(_attributes[key]);
+        }
+      }
+      return { _attributes: ret, _modalityId: modalityId, _type: requestJson._type };
```
swashcap: I would remove this function’s body entirely and use this `return`:

```js
return {
  _attributes: response.rows.reduceRight((memo, row) => {
    const _attributeId = parseInt(row._attribute_id, 10);

    if (memo.some(x => x._attributeId === _attributeId)) {
      return memo;
    }

    return memo.concat({
      _attributeDescription = row.attribute_description,
      _attributeId,
      _attributeLabel = row.attribute_label,
      _modalityId = parseInt(row.modality_id, 10),
      _options: [{
        count: parseInt(row.count, 10),
        optionId: parseInt(row.option_id, 10),
        optionLabel: row.option_label,
      }],
      _parentId = parseInt(requestJson._parent_id, 10),
    });
  }, []).reverse(),
  _modalityId: modalityId,
  _type: requestJSON._type,
};
```

This way, you’re not creating the data structure `_attributes`, which just builds a unique hash map of rows based on `attribute_id`. Using `reduceRight` and checking for existing `_attributeId` matches does the trick, and your data structure is pretty clear.

swashcap: You could also do a `Array#filter` and `Array#map` on `response.rows` to filter out unneeded duplicates, then map the row to the required data structure. `Array#reduceRight` avoids that extra iteration.

ruwang:
Thanks! This does looks much cleaner and more clear.  I need to figure out
how all those functions work :)

ruwang:
Good to know 👍.  with Cory's rewrite, this would not be needed.

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,70 @@
...
+      ELSE CAST(o.option_label as integer)
+    END,
+    o.option_label`;
+
+  return bookshelf.knex.raw(qryStr)
+    .then(response => {
+      const ret = [];
+      const _attributes = {};
+      for (let i = 0; i < response.rows.length; i++) {
+        const attributeId = parseInt(response.rows[i].attribute_id, 10);
```
drewlandis: `attribute_id` is type `numeric(4,0)` in the database.  Since it has type `numeric`, I think that means knex (or bookshelf?) casts it to a string.  Rather than casting each of these back to a number using `parseInt`, I think we should cast the column in the sql query to be an int.  E.g. line 7 should like this: `a.attribute_id::int,`

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,70 @@
...
+        if (_attributes[attributeId] === undefined) {
+          _attributes[attributeId] = {};
+          _attributes[attributeId]._options = [];
+        }
+
+        _attributes[attributeId]._attributeId = attributeId;
+        _attributes[attributeId]._attributeLabel = response.rows[i].attribute_label;
+        _attributes[attributeId]._attributeDescription = response.rows[i].attribute_description;
+        _attributes[attributeId]._modalityId = parseInt(response.rows[i].modality_id, 10);
```
drewlandis: Same as above.  modality_id should be cast to an `int` in the sql query above --> `a.modality_id::int,`

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,70 @@
...
+        _attributes[attributeId]._options.push(
+          {
+            optionId: parseInt(response.rows[i].option_id, 10),
```
drewlandis: option_id has type `text` in the database and it is not always numeric.  I think that we should _not_ call `parseInt` on it, otherwise we may end up with `NaN`.  E.g. `parseInt('foo', 10)` --> `NaN`

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,70 @@
...
+            optionLabel: response.rows[i].option_label,
+            count: parseInt(response.rows[i].count, 10),
```
drewlandis: Same as above. count should be cast to an `int` in the sql query above --> `o.count::int`

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,70 @@
...
+          }
+        );
+        _attributes[attributeId]._parentId = parseInt(requestJson._parent_id, 10);
```
drewlandis: I think that this should be done in the root controller.  What do you think?

ruwang:
Nice catch! 👍 all makes sense, except the last one which I don't quite
understand. Do all the controller requests return the Object with
`_parentId`? Does `_parentId` go with all the query?

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,70 @@
...
+       ON m.modality_id = a.modality_id
+    INNER JOIN dx_options_mv o
+       ON a.attribute_id = o.attribute_id
+    WHERE o.option_id IS NOT NULL `;
+    // if modality_id != -1, then we are calling this function from the GUI
+    //    select only the specified modality_id and DON'T show attributes that
+    //    are not supposed to be displayed to user
+    // else modality_id == -1, then we are calling this function from the API
+    //    select ALL modality_id's and ALL attributes
+  if (modalityId !== '-1') {
```
swashcap: Nice catch. That's correct, code in this branch will never execute.

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
swashcap: Ah, interesting. Does the front end ever expect it to be a number? What exactly happens with these `_options`?

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
swashcap: I’d leave it. We should be able to see if everything needs a `_parentId` and refactor once everything’s merged in.

drewlandis: Sounds good to me 👍

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
drewlandis: @swashcap can you explain why you are using `array.reduceRight(fn, []).reverse()` rather than just `array.reduce(fn, [])`?

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,74 @@
...
+    o.option_label`;
+
+  return bookshelf.knex.raw(qryStr)
+    .then(response => { // eslint-disable-line arrow-body-style
+      return {
+        _attributes: response.rows.reduce((memo, row) => {
+          const _attributeId = row.attribute_id;
+          const option = {
+            count: row.count,
+            optionId: parseInt(row.option_id, 10),
```
drewlandis: option_id will not always be a number.  E.g.:
```
attribute_id: 201
attribute_label: Gender
option_id: Male, Female, Unknown
```
Please remove `parseInt` and/or check if the value is numeric-ish prior to parsing.

packages/steelpenny/src/test/integration/dataexchange.js
```diff
@@ -167,10 +167,39 @@ describe('Dataexchange tests', () => {
...
+        'getFormAttributes',
+        '{ "_parentId": null, "_attributes": [], "_type": "Studies", "_modalityId": 143 }'
+      )
+      .then((response) => {
+        const result = response.data;
+        result.data[0].should.deep.equal({
+          result:
+          {
+            _attributes: [
+              { _attributeId: 22,
```
drewlandis: Please add another test for `attribute_id: 201`.  See above

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -36,9 +36,10 @@ const getFormAttributes = (bookshelf, query) => {
       return {
         _attributes: response.rows.reduce((memo, row) => {
           const _attributeId = row.attribute_id;
+          const optionId = isNaN(row.option_id) ? row.option_id : parseInt(row.option_id, 10);
```
swashcap: I’d suggest changing this to use `Number.isNaN`. [`isNaN` has issues](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#Confusing_special-case_behavior).

packages/steelpenny/src/test/integration/dataexchange.js
```diff
@@ -167,10 +167,342 @@ describe('Dataexchange tests', () => {
...
+          },
+        });
+      });
+    });
+
+    it('Gets form attributes, with text option id', () => { // eslint-disable-line arrow-body-style
+      return apiClient.dataexchange.get(
+        'filter',
+        'getFormAttributes',
+        '{"_parentId":null,"_attributes":[],"_type":"Subjects","_modalityId":142}'
```
swashcap: What do you think about using the `getQueryJson` helper function like the other tests?

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,74 @@
...
+    FROM dx_source_attributes_vw a
+    INNER JOIN mrs_modalities_vw m
+       ON m.modality_id = a.modality_id
+    INNER JOIN dx_options_mv o
+       ON a.attribute_id = o.attribute_id
+    WHERE o.option_id IS NOT NULL `;
+    // if modality_id != -1, then we are calling this function from the GUI
+    //    select only the specified modality_id and DON'T show attributes that
+    //    are not supposed to be displayed to user
+    // else modality_id == -1, then we are calling this function from the API
```
swashcap: This `-1` functionality is not needed, right @drewlandis?

packages/steelpenny/src/lib/controllers/dataexchange.js
```diff
@@ -22,8 +24,8 @@ module.exports.get = {
       allowUnknown: true,
     },
     query: {
-      objType: joi.string().allow(['request']),
-      action: joi.string().allow(['loadFilterableModalities', 'getStatistics']),
+      objType: joi.string().allow(['request', 'filter']),
+      action: joi.string().allow(['loadFilterableModalities', 'getFormAttributes', 'getStatistics']),
```
swashcap: Linter is complaining on this line. Want to shorten it?

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+const getFormAttributes = (bookshelf, query) => {
+  const modalityId = parseInt(query.json._modalityId, 10);
+  let qryStr = `SELECT
+    a.attribute_id::int,
```
swashcap: Linter is complaining of whitespace at the end of line 6 and 7.

packages/steelpenny/src/test/integration/dataexchange.js
```diff
@@ -167,10 +167,342 @@ describe('Dataexchange tests', () => {
   });
 
   describe('Filter controller', () => {
-    // individual controller tests go here
-    it('It should do stuff', () => {
-      const stuff = true;
-      stuff.should.equal(true);
+    it('Gets form attributes, with numeric option id', () => { // eslint-disable-line arrow-body-style
```
swashcap: Linter complains on a few of your changes in this file. Please run:

```shell
node_modules/.bin/eslint src/test/integration/dataexchange.js
```

This should show you the errors.

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
ruwang: Fixed. Interesting that My global eslint with version 2.13 does not complaint, but local eslint with version 2.7 does complaint.

packages/steelpenny/src/test/integration/dataexchange.js
ruwang: Thanks. Fixed.

packages/steelpenny/src/test/integration/dataexchange.js
ruwang: That's a good idea.  I take a look at the helper function, however, none of the properties in the object returned by the function will be used here, and for this test, only 4 properties are used. so I am not sure we need to create another local helper function?

packages/steelpenny/src/lib/controllers/dataexchange.js
ruwang: fixed.

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
drewlandis: @swashcap, correct.  Lines 19-23 are commented out in the original dcwebservices code as well, so I would say we should remove them from here rather than carrying them over.

ruwang: I think @swashcap was referring to the API functionality , not comments?

ruwang:
Test added. Note, I was trying to only test for attribute id 201 by passing
thru that attribute id, however, seems that id was not taken when querying
the database. That means the results will be based on the modality id
passed in.

packages/steelpenny/src/lib/controllers/dataexchange/filter.js
swashcap: Looks like both the comments and the `if (modalityId !== -1)` branch aren’t necessary. I’d drop them to avoid future confusion.

packages/steelpenny/src/test/integration/dataexchange.js
swashcap: Hm, good point. I guess we should leave it for now and refactor the helper function if needed.

swashcap:
@ruwang 🎉 Great work!

PR #28, opened 2016-10-25
Dataexchange cookie

mstone121:
#### Asana
# problem statement

User info is need in some of the sub-controllers.
# solution

Add the cookie back to the request headers.


packages/steelpenny/package.json
```diff
@@ -16,6 +16,7 @@
     "bookshelf": "0.9.5",
     "bookshelf-shield": "2.1.2",
     "boom": "2.8.0",
+    "btoa": "1.1.2",
```
swashcap: `btoa`’s functionality already exists here: https://github.com/MRN-Code/quarterback/blob/dcrerwrite/packages/steelpenny/src/lib/models/User.js#L5-L8. I’d suggest refactoring that so we don’t have the extra dependency.

packages/steelpenny/src/lib/controllers/dataexchange.js
```diff
@@ -25,28 +26,39 @@ module.exports.get = {
       objType: joi.string().allow(['request']),
       action: joi.string().allow(['loadFilterableModalities', 'getStatistics']),
       json: joi.string(),
+      cookies: joi.string()
     },
   },
   handler: (clientRequest, reply) => {
-    let query;
+    let query, cookie;
```
swashcap: Declaring multiple variables per line isn’t allowed in our linting rules.

swashcap: I’d run all this through the linter (`node_modules/.bin/eslint` from the repo root). Seeing multiple things that need to be fixed.

packages/steelpenny/src/lib/controllers/dataexchange.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 const boom = require('boom');
 const joi = require('joi');
+const ccu = require('../utils/cas-cookie-utils.js');
```
swashcap: It’s clearer to refer to the module by it’s full name (`casCookieUtils`). This makes subsequent method calls easier to understand.

packages/steelpenny/src/lib/controllers/dataexchange.js
```diff
@@ -25,28 +26,39 @@ module.exports.get = {
...
-    .then(data => reply({
-      result: data,
-    }))
-    .catch(err => reply(boom.badData(err)));
+    function callSubController() {
+      return subControllers[clientRequest.query.objType]
+        [clientRequest.query.action](clientRequest.server.plugins.bookshelf, query, cookie);
+    }
+
+    Promise.resolve(parseCookie)
```
swashcap: `parseCookie` is already a Bluebird-y `Promise.` No need to wrap it with `Promise.resolve`.

I’d also suggest changing code order to improve legibility. `parseCookie` is only used once, and it’s declared far from where it’s used. Instead, why not use `casCookieUtils.verifyAndParse` directly and chain onto it:

```js
casCookieUtils.verifyAndParse(
  clientRequest.query.cookies,
  clientRequest.server.plugins['hapi-redis'].client
)
  .then(/*...*/)
  //...
```

Futhermore, hapi’s `reply` should only be called once. Your Promise chain could result in it being called _three times_. It’s also difficult to track what's happening to the `cookie` variable as it’s assigned far from where it's declared, and parsed. This should be fixed.

mstone121: I did this on purpose. I wanted the actual promise change to be readable, so I wrapped each step in a function that describes what it does (so each could be on one line). As far as distance between parseCookie and where it's used. I essentially just followed hoisting rules and declared the functions in the order they are called so I think it's pretty readable. 

I think the best way to make `cookie` make sense would be to pass it along the chain rather then declaring it in the outside scope. 

Is there a way to break the chain when catching errors? That's what I intended to happen and I'm not sure how I would make error messages distinct otherwise.

packages/steelpenny/package.json
swashcap: The `username` and `password` fields are only base64-encoded. We don't need to decrypt them, right?

//cc @rssk

swashcap:
> I've fixed the linting errors. What is the reasoning behind including the trailing comma?

Nice, looks good. I believe the reasoning behind the trailing comma is that it’s less work to add properties to an object or items to an array, thus making cleaner diffs.

packages/steelpenny/src/lib/controllers/dataexchange.js
swashcap: This is too confusing. Instead of creating functions that wrap functions, just call 'em directly. This is how I’d write it:

```js
module.exports.get = {
  // ...
  handler(
    {
      query,
      query: {
        action,
        cookies,
        objType,
      },
      server: { plugins },
    },
    reply
  ) {
    const doReply = once(reply);

    return casCookieUtils.verifyAndParse(cookies, plugins['hapi-redis'].client)
      .catch(error => doReply(boom.unauthorized('Could not parse cookie', error)))
      .then(cookie => {
        let parsedQuery;

        if (query.json) {
          try {
            parsedQuery = Object.assign({}, query, {
              json: JSON.parse(query.json),
            });
          } catch (error) {
            return doReply(boom.badData('Could not parse JSON', error));
          }
        } else {
          parsedQuery = query;
        }

        return subControllers[objType][action](
          plugins.bookshelf,
          parsedQuery,
          cookie
        );
      })
      .then(data => doReply({ result: data }))
      .catch(error => doReply(boom.badImplementation(error)));
  },
};
```

Writing in this manner makes the control flow easier to read: variable declaration and assigmnent isn’t spread throughout the function body; wrapper functions aren’t declared and called with params from multiple scopes. Destructuring the arguments makes it a bit easier to understand their shape, and it makes their use in `subControllers` a bit easier. I’d also use [`lodash#once`](https://lodash.com/docs/4.16.4#once) to wrap `reply` and ensure it's only ever called once.

swashcap: > Is there a way to break the chain when catching errors?

Yeah, I’m not sure of the best way to do this. [`Bluebird.reflect`](http://bluebirdjs.com/docs/api/reflect.html) could work, or you could use [custom errors like we're doing in penny-collector_](https://github.com/MRN-Code/penny-collector/blob/a893cd9d784c16dc29bcbd08c0b0565287f55cb6/src/PennyCollector.js#L41-L52).

mstone121: The problem with reflect is that db changes would happen before the error is caught. 

With the custom error handler, would that call `reply`? If not I'm not sure server error responses will be at all helpful.

packages/steelpenny/package.json
```diff
@@ -55,6 +54,7 @@
     "pg": "6.0.0"
   },
   "devDependencies": {
+    "abab": "1.0.3",
```
swashcap: Why did this get moved to `devDependencies`? We depend on it here: https://github.com/MRN-Code/quarterback/blob/6125fd502f6ae65a2f8b8801e21840832a9a8756/packages/steelpenny/src/client/authClient.js#L9-L10

rssk: I requested that, authclient is not used by the application, only on builds and for testing

swashcap: Perhaps a bad merge? Fixed in f703858904a62cc2edf9dbed70df719825ace288.

PR #35, opened 2016-11-07
Refactor `/auth/keys` route

swashcap:
* **Problem:** The _/auth/keys_ route is kinda messy
* **Solution:** Refactor!
* **Testing:** Run the tests!


packages/steelpenny/src/lib/controllers/keys.js
```diff
@@ -147,49 +136,45 @@ module.exports.delete = {
...
+      },
+      params: {
+        id: paramsId,
+      },
+      server: { plugins },
+    } = request;
+    const redisClient = plugins['hapi-redis'].client;
+
+    if (id !== paramsId) {
+      throw boom.unauthorized('Access to this key not allowed');
```
ruwang: Note: might need to investigate. using `return reply()` instead of `throw`

packages/steelpenny/src/lib/controllers/keys.js
```diff
@@ -147,49 +136,45 @@ module.exports.delete = {
...
+        // delete the API key
+        return Promise.all([
+          redisClient.del(id),
+          redisClient.srem(username, id),
+        ]);
+      })
+
+      // reply with success message and new invalid cookie
+      .then(() => reply(internals.logOutSuccessObj)
+        .code(200)
```
rssk: May not be necessary? I think reply should default to 200, but there is extra state being set, so not sure

packages/steelpenny/src/lib/controllers/keys.js
```diff
@@ -147,49 +136,45 @@ module.exports.delete = {
...
+        .state(casCookieUtils.cookieName, casCookieUtils.invalidate())
+      )
+      .catch(error => {
+        request.log(['error', 'logout'], error);
+        return reply(boom.wrap(error));
```
rssk: Not looking for boom errors here, however this would probably require a refactor of how cookie/auth utils work

packages/steelpenny/src/lib/controllers/keys.js
swashcap: Fixed in 1543ee25ae8b914fbd722cac49dcb3815fb6b861.

packages/steelpenny/src/lib/controllers/keys.js
swashcap: I assume `boom#wrap` handles boom errors responsibly.

packages/steelpenny/src/lib/controllers/keys.js
swashcap: Tests still pass when the `.code(200)` line is dropped. So…?

packages/steelpenny/src/lib/controllers/keys.js
rssk: I think so, what I'm saying is that this doesnt follow the paradigm of only reply w/ errors we can handle, however I don't think there is an easy fix here for that.

packages/steelpenny/src/lib/controllers/keys.js
rssk: They'd fail if the response code changed, `kill it`

swashcap: Killed in f176e429f2feca5c02d83c8e00d979df15546d00.

PR #53, opened 2016-11-15
Add joi schema to sites controller

swashcap:
* **Problem:** The `/sites` route has no documented return value
* **Solution:** Add a joi schema!
* **Testing:**
  1. Spin up a server
  2. Run some requests through the _/swagger/documentation_ site getter
  3. Confirm joi doesn't explode


swashcap:
While this isn’t necessary for the PR site, MRN-Code/coins-pr-site@ 929e804104fb18c35a5f080ecda5e0a3c3da7f6a was based off this feature.

swashcap:
39decbfa23e41144fff089e8d4026718eff78a5d fixes tests. 💚

packages/steelpenny/src/lib/controllers/sites.js
```diff
@@ -45,7 +82,10 @@ module.exports.get = {
     },
   },
   response: {
-    schema: internals.siteSchema,
```
drewlandis: interesting... `internals.siteSchema` was resolving to `undefined` thus allowing _anything_ through.  Then, when you specified a valid joi schema, it failed because the response is sometimes of the form:
```
{schema}
```
and other times of the form:
```
[{schema}, {schema}, {schema}]
```
This is out of the purview of this PR, but I'm wondering if the site route logic should be updated so that the response always has a consistent form.  E.g. even when only returning a single response, it should have the form: `[{schema}]`

PR #54, opened 2016-11-15
Dataexchange group delete

jwtlake:
## Overview
Added group delete sub-controller and tests.

## Usage 
This route is used when the user deletes a group from the request.

## Outcome
The sub controller takes the provided group id and executes a DB function that deletes the group, any child groups, and any filters. 
**Returns:** If successful it returns a empty object.
```javascript 
{
  result: {
     _attributes: [],
  },
}
```
## DBFunction
**Name:** dx_delete_data_request_template_group_f(groupID, Username)
**OnSuccess:** returns the groupId it just deleted 
**OnFailure:** throws a sql error

The function will error out if the request has already been submitted (so it's frozen) or if the user doesn't have access rights.


packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +105,46 @@ const switchType = (bookshelf, query, cookie) => {
     });
 };
 
+/**
+ * Executes a DB function that deletes a group and all it's childern
```
swashcap: `it's` → `its`

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +105,46 @@ const switchType = (bookshelf, query, cookie) => {
...
+  const username = cookie.username;
+
+  const funcStr = 'SELECT dxdba.dx_delete_data_request_template_group_f(?,?);';
+  const prams = [id, username];
+
+  return bookshelf.knex.raw(funcStr, prams)
+  .then((response) => {
+    // verify that the function deleted the correct Group_id
+    // TODO: This check might not be necessary
+    const responseId = parseInt(response.rows[0].dx_delete_data_request_template_group_f, 10);
```
swashcap: Wrap the `response.rows[0].dx_delete_data_request_template_group_f` in a [`lodash#get`](response.rows[0].dx_delete_data_request_template_group_f) call such that you don’t get a bad object lookup error.

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +105,46 @@ const switchType = (bookshelf, query, cookie) => {
...
+      throw boom.badImplementation(
+        'result error',
+        'Delete function returned a differnt Group_id then what it was called with.'
+        + ` Called: ${id} Returned: ${responseId}`
+      );
+    }
+    // return empty object
+    return { _attributes: [] };
+  })
+  .catch((error) => {
```
swashcap: Let the root handler (dataexchange controller) handle rejections. Drop the `.catch` here.

packages/steelpenny/test/integration/dataexchange.js
```diff
@@ -1273,10 +1273,131 @@ describe('Dataexchange tests', () => {
...
+          const expected = {
+            result: {
+              _attributes: [],
+            },
+          };
+          result.data[0].should.deep.equal(expected);
+        });
+      });
+
+      it('throws an error when user doesnt have rights', () => {
```
swashcap: Change the “user doesnt have rights” to something like “user doesn't have access” or “user doesn't have permission” to clarify the test's purpose.

packages/steelpenny/test/unit/dataexchange.js
```diff
@@ -4,10 +4,20 @@ const chai = require('chai');
...
+const group = require('../../src/lib/controllers/dataexchange/group');
 
 // Set should property of all objects for BDD assertions
 chai.should();
 
+// helper function for mocking up bookshelf responses
+const bookShelfOverride = (returnObj) => { // eslint-disable-line arrow-body-style
+  return {
+    knex: {
+      raw: () => new Promise((fulfill) => fulfill(returnObj)),
```
swashcap: Please use `Promise#reject` here:

```diff
-      raw: () => new Promise((fulfill) => fulfill(returnObj)),
+      raw: () => Promise.reject(returnObj),
```

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +105,46 @@ const switchType = (bookshelf, query, cookie) => {
...
+
+  return bookshelf.knex.raw(funcStr, prams)
+  .then((response) => {
+    // verify that the function deleted the correct Group_id
+    // TODO: This check might not be necessary
+    const responseId = parseInt(response.rows[0].dx_delete_data_request_template_group_f, 10);
+    if (responseId !== id) {
+      throw boom.badImplementation(
+        'result error',
+        'Delete function returned a differnt Group_id then what it was called with.'
```
swashcap: Please correct spelling of “differnt” throughout these file changes, and update “then” to “than”.

You can also use [string templating](https://github.com/airbnb/javascript#strings--line-length) here, too.

packages/steelpenny/test/unit/dataexchange.js
```diff
@@ -52,7 +62,27 @@ describe('Dataexchange tests', () => {
   });
 
   describe('Group controller', () => {
+    describe('GET delete', () => {
+      it('Returns an error when the delete function returns a differnt id then expected', () => {
+        const returnObj = { rows: [{ dx_delete_data_request_template_group_f: 999 }] };
```
swashcap: Your test doesn’t use `returnObj` is only used once. Just throw the object in the `bookShelfOverride` call.

packages/steelpenny/test/unit/dataexchange.js
```diff
@@ -52,7 +62,27 @@ describe('Dataexchange tests', () => {
...
+        const cookie = { username: 'mochatest' };
 
+        return group.delete(bookshelf, query, cookie)
+        .catch((error) => {
+          error.isBoom.should.equal(true);
+          error.output.statusCode.should.equal(500);
+          error.output.payload.error.should.equal('Internal Server Error');
+          error.message.should.equal('result error');
+          error.data.should.equal(
+            'Delete function returned a differnt Group_id then what it was called with.'
```
swashcap: Same thing with string templating here.

packages/steelpenny/test/unit/dataexchange.js
```diff
@@ -52,7 +62,27 @@ describe('Dataexchange tests', () => {
...
+
+        const bookshelf = bookShelfOverride(returnObj);
+        const query = { json: { _id: '1001' } };
+        const cookie = { username: 'mochatest' };
 
+        return group.delete(bookshelf, query, cookie)
+        .catch((error) => {
+          error.isBoom.should.equal(true);
+          error.output.statusCode.should.equal(500);
+          error.output.payload.error.should.equal('Internal Server Error');
```
swashcap: This line tests boom's internal error strings. Just test for a `500` error. In fact, you might be able to drop this part of the unit test if you drop `group.delete`’s `.catch` as suggested above.

packages/steelpenny/src/lib/controllers/dataexchange/group.js
jwtlake: fixed

packages/steelpenny/src/lib/controllers/dataexchange/group.js
jwtlake: fixed spelling and changed to single line string templating.

packages/steelpenny/test/integration/dataexchange.js
jwtlake: updated

packages/steelpenny/test/unit/dataexchange.js
jwtlake: updated

packages/steelpenny/test/unit/dataexchange.js
jwtlake: Do I need a string template here? There are no variables. I moved it onto one line but now I'm over the 100 char limit. Not sure what I need to do for these.

swashcap: Just add `// eslint-disable-line max-len` until we roll in the new AirBnB style guide.

packages/steelpenny/src/lib/controllers/dataexchange/group.js
jwtlake: So I kinda think we still want to keep this here because of the way the function was written. (Throwing a SQL error vs returning a value)

If I remove this catch then the server would crash every time a user tried to delete a filter or group when they didn't have permission to do so.

rssk: Then throw an appropriate boom error (forbidden, perm denied, etc...)

jwtlake: Updated catch logic

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -125,7 +125,7 @@ const deleteById = (bookshelf, query, cookie) => {
   .then((response) => {
     // verify that the function deleted the correct Group_id
     // TODO: This check might not be necessary
-    const responseId = parseInt(get(response, 'rows[0].dx_delete_data_request_template_group_f'), 10);
+    const responseId = parseInt(get(response, 'rows[0].dx_delete_data_request_template_group_f'), 10); // eslint-disable-line max-len
```
swashcap: Please don't disable `max-len`. Wrap the line.

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -136,10 +136,10 @@ const deleteById = (bookshelf, query, cookie) => {
     return { _attributes: [] };
   })
   .catch((error) => {
-    if (error.isBoom) {
-      throw error;
+    if (error.message.startsWith('Cannot delete a group from a submitted request')) {
```
swashcap: Clarify with a comment where this error message string is coming form. It isn't immediately clear to those not familiar with the code.

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -136,10 +136,10 @@ const deleteById = (bookshelf, query, cookie) => {
...
+      throw boom.forbidden('submitted request', error.message);
```
swashcap: Just pass the entire `error` for now until we have time to examine boom’s internals.

jwtlake:
I attempted a `git commit -n —amend` on the last commit which seems to have worked but now I have an additional weird merge right after that has no change files. bahh. I have no idea what I’m doing…

packages/steelpenny/lib/controllers/dataexchange/group.js
```diff
@@ -102,8 +104,52 @@ const switchType = (bookshelf, query, cookie) => {
       return { _children: [], _filters: [] };
     });
 };
+/* eslint-disable max-len */
```
swashcap: Don’t ignore maximum length for doc blocks. Please wrap parameter descriptions like so:

```diff
- * @param  {Object}  cookie  User's cookie object which includes thier username for permission verification
+ * @param  {Object}  cookie  User's cookie object which includes thier username for
+ * permission verification
```

packages/steelpenny/lib/controllers/dataexchange/group.js
```diff
@@ -102,8 +104,52 @@ const switchType = (bookshelf, query, cookie) => {
...
+      throw boom.badImplementation(
+        'result error',
+        `Delete function returned a different Group_id than what it was called with. Called: ${id} Returned: ${responseId}` // eslint-disable-line max-len
+      );
+    }
+    // return empty object
+    return { _attributes: [] };
+  })
+  .catch((error) => {
+    // check the error message to see if it matches a RAISE EXCEPTION error returned from the db
```
swashcap: 👍

packages/steelpenny/lib/controllers/dataexchange/group.js
```diff
@@ -102,8 +104,52 @@ const switchType = (bookshelf, query, cookie) => {
       return { _children: [], _filters: [] };
     });
 };
+/* eslint-disable max-len */
+/**
+ * Executes a DB function that deletes a group and all its childern
+ *
+ * @param  {Object}  bookshelf  Hapi ORM pluggin used to make DB calls
```
swashcap: You don’t need the multiple spaces here. JSDoc can parse it fine like so:

```diff
- * @param  {Object}  bookshelf  Hapi ORM pluggin used to make DB calls
+ * @param {Object} bookshelf Hapi ORM pluggin used to make DB calls
```

packages/steelpenny/test/integration/dataexchange.js
```diff
@@ -1273,10 +1273,130 @@ describe('Dataexchange tests', () => {
...
+              ],
+              _statistics: '1',
+              _type: 'Studies',
+              _modalityId: 143,
+            },
+          ],
+          _id: 1, // group id user doesnt have access to
+          _children: [],
+          _type: 'and',
+        };
```
swashcap: Please add a line break here to separate the data structure from the code that’s running the test.

packages/steelpenny/test/unit/dataexchange.js
```diff
@@ -52,7 +62,21 @@ describe('Dataexchange tests', () => {
   });
 
   describe('Group controller', () => {
+    describe('GET delete', () => {
+      it('Returns an error when the delete function returns a different id than expected', () => {
```
swashcap: Your method isn’t _returning an error_, it’s _rejecting_. Please update the description.

packages/steelpenny/test/unit/dataexchange.js
```diff
@@ -52,7 +62,21 @@ describe('Dataexchange tests', () => {
...
+        const bookshelf = bookShelfOverride({ rows: [{ dx_delete_data_request_template_group_f: 999 }] }); // eslint-disable-line max-len
```
swashcap: Please don’t ignore max length on this line.

packages/steelpenny/lib/controllers/dataexchange/group.js
jwtlake: Ahhh.. Haha now I feel silly. Thanks
Fixed!

packages/steelpenny/test/unit/dataexchange.js
jwtlake: Updated description

jwtlake:
Rebased with dcrerwrite to cleanup commit history and get latest changes. I'd say this is ready for review except for the note below.

**Note:** A test is failing for the new generateSourceObject changes. This is happening for me in the latest dcrerwrite branch as well.
```1) Dataexchange tests Request controller GET generateSourceObject Generates a source object:```

swashcap:
:tada: Yeah! Nice work.

PR #59, opened 2016-11-16
Dataexchange getfilterstatistics

ruwang:
## Asana
https://app.asana.com/0/8755179914462/210932248794574

# Feature
Add group's load filter statistics module for dc rewrite


packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +103,60 @@ const switchType = (bookshelf, query, cookie) => {
     });
 };
 
+/**
+ * This will get the statistics (result count) for each filter and add this statistics back
+ * to the original query object and return the modified object.
+ *
+ * @param {Object} bookshelf
+ * @Param {Object} query    Json tree.
```
swashcap: Lowercase “Param” here.

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +103,60 @@ const switchType = (bookshelf, query, cookie) => {
...
+    ),
+    Promise.all(
+      filters.map(f => {
+        const filter = clone(f);
+        const attribute = filter._attributes[0];
+        const attributeId = attribute._attributeId;
+        const optionId = attribute._optionId;
+        const optionLabel = attribute._optionLabel;
+
+        if (optionLabel === 'USER_INPUT') {
```
swashcap: Please add a comment explaining why we shouldn’t filter when in `USER_INPUT` mode. This should make sense for developers unfamiliar with the code/db functions.

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +103,60 @@ const switchType = (bookshelf, query, cookie) => {
...
+  const childGroups = group._children;
+  const filters = group._filters;
+
+  // Recursively get the statistics for each filter in the query
+  return Promise.all([
+    Promise.all(
+      childGroups.map(childGroup => loadFilterStatistics(bookshelf, childGroup))
+    ),
+    Promise.all(
+      filters.map(f => {
```
swashcap: Change `f` to something more descriptive.

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +103,60 @@ const switchType = (bookshelf, query, cookie) => {
...
+
+        return bookshelf.knex.raw(qryStr)
+          .then((rs) => {
+            filter._statistics = parseInt(rs.rows[0].count, 10);
+            return filter;
+          });
+      })
+    ),
+  ])
+  .then(([returnedGroups, returnedFilters]) => {
```
swashcap: What do you think about writing this last `.then` block like so?

```js
.then(([returnedGroups, returnedFilters]) => Object.assign({}, group, {
  _children: returnedGroups,
  _filters: returnedFilters,
});
```

Even though `group` is `clone`-ed up top, this doesn't mutate it and makes the return a little easier to read.

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +103,60 @@ const switchType = (bookshelf, query, cookie) => {
...
+  const filters = group._filters;
+
+  // Recursively get the statistics for each filter in the query
+  return Promise.all([
+    Promise.all(
+      childGroups.map(childGroup => loadFilterStatistics(bookshelf, childGroup))
+    ),
+    Promise.all(
+      filters.map(f => {
+        const filter = clone(f);
```
rssk: Since you're cloning on 115, this clone down here is unnecessary, cloning should only really be used so your function `loadFilterStatistics` doesn't modify the passed in objects, map is an internal function, so that's fine. You can use the eslint block disable to prevent it barking :-)

packages/steelpenny/src/lib/controllers/dataexchange/group.js
```diff
@@ -103,7 +103,60 @@ const switchType = (bookshelf, query, cookie) => {
...
+        }
+        const qryStr = `
+          SELECT count
+          FROM dx_options_mv
+          WHERE attribute_id = ${attributeId}
+            AND option_id = '${optionId}';
+        `;
+
+        return bookshelf.knex.raw(qryStr)
+          .then((rs) => {
```
swashcap: I’d avoid nested Promise chaining. Here’s how I’d write this function to avoid nesting and cloning altogether:

```js
const loadFilterStatistics = (bookshelf, query) => {
  const group = query.json || query;
  const {
    _children: childGroups,
    _filters: filters,
  } = group;

  // Recursively get the statistics for each filter in the query
  return Promise.all([
    Promise.all(
      childGroups.map(childGroup => loadFilterStatistics(bookshelf, childGroup))
    ),
    Promise.all(filters.map(filter => {
      const {
        _attributes: [{
          _attributeId: attributeId,
          _optionId: optionId,
          _optionLabel: optionLabel,
        }],
      } = filter;

      return Promise.all([
        filter,
        optionLabel === 'USER_INPUT' ?
          undefined :
          bookshelf.knex.raw(`
            SELECT count
            FROM dx_options_mv
            WHERE attribute_id = ${attributeId}
              AND option_id = '${optionId}';
          `),
      ]);
    })),
  ])
  .then(([groups, filtersObjs]) => Object.assign({}, group, {
    _children: groups,
    _filters: filtersObjs.map(
      ([filter, response]) => Object.assign({}, filter, {
        _statistics: typeof response === 'object' ?
          parseInt(response.rows[0].count, 10) :
          '',
      })
    ),
  }));
};
```

_(Not tested!)_

swashcap: This might be more difficult to read due to that nested `Promise.all([filter, /* other stuff */])`. You operate on `filtersObjs` depending on that array’s second item, which is disconnected from the bookshelf call. 🤔

ruwang: I like this 👍

packages/steelpenny/test/integration/dataexchange.js
```diff
@@ -1273,10 +1274,241 @@ describe('Dataexchange tests', () => {
...
+
+          return apiClient.dataexchange.get(
+            'group',
+            'loadFilterStatistics',
+            JSON.stringify(queryJson),
+            cookie
+          )
+          .then((response) => {
+            const result = response.data;
+            const expectedJson = clone(queryJson);
```
swashcap: `clone` isn’t necessary here because you’re `JSON.stringify`-ing up above.

packages/steelpenny/test/integration/dataexchange.js
```diff
@@ -1273,10 +1274,241 @@ describe('Dataexchange tests', () => {
...
+                    _modalityId: 112,
+                  },
+                ],
+                _id: 988,
+                _children: [],
+                _type: 'or',
+              },
+            ],
+            _type: 'and',
+          };
```
swashcap: Might be nice to add a <del>space</del> <ins>line break</ins> after this line to separate the test data structure from the actual test code.

packages/steelpenny/test/integration/dataexchange.js
```diff
@@ -1273,10 +1274,241 @@ describe('Dataexchange tests', () => {
...
+          return apiClient.dataexchange.get(
+            'group',
+            'loadFilterStatistics',
+            JSON.stringify(queryJson),
+            cookie
+          )
+          .then((response) => {
+            const result = response.data;
+            result.data[0].should.deep.equal({
```
swashcap: Might be nice to reduce duplicate data structures here. You could define `queryJson` as this test structure here, then strip every `_statistics` property when passing it into `apiClient.dataexchange.get`. So:

```js
const expectedResult = { /* ... */ }

return apiClient.dataexchange.get(
  'group',
  'loadFilterStatistics',
  JSON.stringify(stripEveryStatisticsProperty(expectedResult)),
  cookie
)
  .then(({ data: [response] }) => {
    response.should.deep.equal(expectedResult);
  });
```

Does that make any sense?

(`stripEveryStatisticsProperty` should be easy to create with some [lodash](https://lodash.com/docs) helpers.)

packages/steelpenny/test/integration/dataexchange.js
```diff
@@ -1311,18 +1311,28 @@ describe('Dataexchange tests', () => {
...
+              result: queryJson,
             });
           });
         }
       );
 
       it('Load filter statistics without USER_INPUT as option label',
         () => {
-          const queryJson = {
+          const removeFilterStat = (group) => {
```
swashcap: 👍

swashcap:
@ruwang This is ready to merge, but there’s some conflicts. Please resolve, then I’ll green light it.

ruwang:
@swashcap Thanks! dcrerwrite merged into this PR. I did not see conflict during merge though, not sure why.

swashcap:
:clap: Nice work!

PR #62, opened 2016-11-21
update shield, remove proxy for clean npm tree

rssk:
# problem statement
- Publishing is failing, storage proxy and BK-shield are the issue
# solution
- bump shield, 👋 proxy, for now.


swashcap:
🎉 Nice!

PR #70, opened 2016-11-28
Fix generateSourceObject Test

drewlandis:
#### asana
_none_

# problem statement
The test for `generateSourceObject` was failing because it is attempting to reference a `data_request_id` that does not yet exist.

# solution
Update the above mentioned test so that it instead references an old data_request_id


swashcap:
🍾 Looks good!

PR #72, opened 2016-11-29
tus child process fix

jwtlake:
# high level problem
npm test was failing with a port in use error
`Error: listen EADDRINUSE 0.0.0.0:7113`

# background
One of the things this branch adds is a [tus server](https://github.com/tus/tus-node-server) implementation for file uploads. The way the tus server is launched is with a child process [here](https://github.com/MRN-Code/quarterback/blob/pennyfile-migration/packages/steelpenny/src/index.js#L29).

# specific problem
The [test script](https://github.com/MRN-Code/quarterback/blob/pennyfile-migration/packages/steelpenny/package.json#L87) calls [build](https://github.com/MRN-Code/quarterback/blob/pennyfile-migration/packages/steelpenny/package.json#L86) prior to running all tests. Part of the build process starts up HP which then starts a tus server.

It seems that the child process (tus) isn't always exiting when the parent process (HP) exits. The build process will complete but the tus server is still running. Then when our actual tests start to run we get the port in use error because HP is trying to start up a new Tus server and the port is already in use.

# solution
This might be a bad idea but I got around the issue by adding a process exit hook to kill the tus server when the parent exits. See changes.


jwtlake:
open sesame!

PR #74, opened 2016-11-29
Added hook to kill tus server on process exit

jwtlake:
# high level problem
npm test was failing with a port in use error
`Error: listen EADDRINUSE 0.0.0.0:7113`

# background
One of the things this branch adds is a [tus server](https://github.com/tus/tus-node-server) implementation for file uploads. The way the tus server is launched is with a child process [here](https://github.com/MRN-Code/quarterback/blob/pennyfile-migration/packages/steelpenny/src/index.js#L29).

# specific problem
The [test script](https://github.com/MRN-Code/quarterback/blob/pennyfile-migration/packages/steelpenny/package.json#L87) calls [build](https://github.com/MRN-Code/quarterback/blob/pennyfile-migration/packages/steelpenny/package.json#L86) prior to running all tests. Part of the build process starts up HP which then starts a tus server. It seems that the child process (tus) isn't always exiting when the parent process (HP) exits. The build process will complete but the tus server is still running. Then when our actual tests start to run we get the port in use error because HP is trying to start up a new tus server and the port is already in use.

# solution
This might be a bad idea but I got around the issue by adding a process exit hook to kill the tus server when the parent exits. See changes.


packages/steelpenny/src/index.js
```diff
@@ -26,7 +26,10 @@ server.app.pluginsRegistered = plugins.reduce(
 .then(() => {
   server.start(() => {
     server.log(['startup'], `server running at: ${server.info.uri}`);
-    childProcess.fork('./src/tus/tus-server.js');
+
+    // start tus server
+    const tus = childProcess.fork('./src/tus/tus-server.js');
+    process.on('exit', () => tus.kill()); // kill tus server on exit
```
rssk: You'll need to call exit again, as you've interrupted it. Also, this may not be the only signal you'll need to listen for.

jwtlake: Oh weird.. So you're saying that listening for exit this way will act like a catch and actually stop the exit? 

If thats true then wouldn't adding another process exit just get caught again by the same function?

swashcap: The nested `process.on` listener can be a bit confusing due to its location. There might be a way to instead add this listener to the server’s `stop` or `internal-error` events, which seems more appropriate. This works for now, but ideally Tus is loaded as a hapi plugin.

drewlandis:
Here is what Ross and my solution looked like.  There may need to be more work that is done still.
`steelpenny/src/index.js`
```diff
-    childProcess.fork('./src/tus/tus-server.js');
+    const tus = childProcess.fork('./src/tus/tus-server.js');
+
+    // kill tus server on user entered: ctrl+c
+    process.on('SIGINT', () => {
+      tus.kill();
+      process.exit(0);
+    });
+
+    // kill tus server on steel penny server exit
+    process.on('exit', () => {
+      tus.kill();
+      process.exit(0);
+    });
```

swashcap:
According to [the `process.on('exit')` docs](https://nodejs.org/api/process.html#process_event_exit):

> There is no way to prevent the exiting of the event loop at this point, and once all 'exit' listeners have finished running the Node.js process will terminate.

So, you won't need those `process.exit(0)` calls. Additionally, it looks like Node.js maps `SIGINT` to a `exit` listeners calls. So, something like:

```
const stopTus = () => tus.kill();

process.on('exit', stopTus);
process.on('SIGTERM', stopTus);
```

PR #79, opened 2016-12-01
added missing attribute key

jwtlake:
# problem statement
failing user tests

# solution
fixed a couple nested key-value lookups that were missing object properties


drewlandis:
👍

PR #88, opened 2016-12-15
Refactor authentication utility

swashcap:
* **Problem:** The authentication utility is difficult to reason about.
* **Solution:** Refactor it!
* **Testing:**
  * Ensure it passes linting: `node_modules/.bin/eslint src/lib/utils/authentication.js`
  * Ensure tests pass on Node.js@4


jwtlake:
Tests pass on node v4.4.7 and v6.9.2

packages/steelpenny/src/lib/utils/authentication.js
```diff
@@ -1,287 +1,264 @@
...
+     * @param {string} id
+     * @param {Function} [callback]
+     * @returns {Promise}
      */
-  authUtils.hashPassword = (password) => {
-    return bcryptGenSalt(12)
-            .then(_.partial(bcryptHash, password));
-  };
+    getHawkCredentials(id, callback) {
+      const cb = typeof callback === 'function' ? callback : noop;
```
swashcap: Correct!

packages/steelpenny/src/lib/utils/authentication.js
```diff
@@ -1,287 +1,264 @@
...
+     *
+     * @param {Bookshelf.Model} user User model
+     * @return {Promise<Object>} Credentials object
      */
-  authUtils.compareHash = (secret, hash) => {
-    return bcryptCompare(secret, hash);
-  };
+    generateHawkCredentials(user) {
+      const username = user.get('username');
+      const encryptedUsername = authUtils.encryptUsername(username);
```
swashcap: Good point: this functionality belongs to a user. Let’s do that on a separate PR to keep this one focused.

packages/steelpenny/src/lib/utils/authentication.js
```diff
@@ -1,287 +1,264 @@
...
+        .then((roles) => Promise.all([
+          roles,
+          revokeRoles(username),
+        ]))
+        .then((responses) => {
+          // TODO: Use desctructuring on args when moving to Node.js >=6
+          const roles = responses[0];
+          const revoked = responses[1];
+
+          return addRolesFromCollection(username, roles, revoked);
```
swashcap: Hmm. Good point, doesn’t look like its used. I wonder why the `revokeRoles` response was being used in the first place.

packages/steelpenny/src/lib/utils/authentication.js
```diff
@@ -1,287 +1,264 @@
...
+            );
 
-        // TODO: if User is ever protected with the bookshelf shield,
-        // we will need to figure out a way to get around that...
-    return new User({ username: encryptedUsername })
-            .fetch()
-            .then(checkAccountStatus)
-            .then(comparePassword)
-            .then(handleCompare);
+            return Promise.resolve(false);
```
swashcap: I imagine `validateUser`’s callers expect a Promise that resolves to `false` to indicate unsuccessful validation.

swashcap:
> This whole process is really confusing to me.

Yeah, doesn’t make 100% sense to me, either. At least this PR makes our authentication operations a bit clearer.

PR #91, opened 2016-12-19
update Readme.md to include info about how halfpenny is generated from steelpenny

drewlandis:
#### asana
https://app.asana.com/0/8755179914444/232497019191795

# Note
This was the best I could figure out how things are working together. Please look this over and make sure that my interpretation is accurate to how things are actually working.

# problem statement
It is confusing how halfpenny is generated...

# solution
...add info to the README that will help clarify


swashcap:
Great writeup!

PR #93, opened 2016-12-20
Refactor environment values

swashcap:
* **Problem:** We’re referencing environment variables and constructing env-based strings in _so many places_:
  * PR site: [new Webpack config](https://github.com/MRN-Code/coins-pr-site/blob/3fb8c2f5e6c6c12e1d254a2cea7497498da6d5a2/webpack.config.js#L35-L51)
  * _coins_core_: [here](https://github.com/MRN-Code/coins_core/blob/f508ca6d89d6300379fb25237a2c5fb0ac90f459/js/browserApp/utils/get-nodeapi-baseurl.js) and [here](https://github.com/MRN-Code/coins_core/blob/f508ca6d89d6300379fb25237a2c5fb0ac90f459/js/browserApp/webpack.config.js#L9)
  * steelpenny: [cookie name](https://github.com/MRN-Code/quarterback/blob/b34665a704c5ea9b70ecbce3e34308d266df0f04/packages/steelpenny/src/lib/utils/cas-cookie-utils.js#L23), [cli “mode”](https://github.com/MRN-Code/quarterback/blob/b34665a704c5ea9b70ecbce3e34308d266df0f04/packages/steelpenny/src/lib/utils/cli-options.js#L39), etc.
  * [coins-grunt-env](https://github.com/MRN-Code/coins-grunt-env), which is required by a few 2.0 repos
  * ???
* **Solution:** Refactor environment-specific values to be set in one location, a new monorepo package named _deposit-box_. This sets:
  * `authCookie`: Name of authentication cookie
  * `env`: Value of environment (one of `production`, `staging`, `development`, `ops` or `test`)
  * `apiURL`: URL for API server, with base endpoint included
  * `webURL`: URL for main site (PR Site)

PR #96, opened 2016-12-27
Drop validate

swashcap:
* **Problem:** Validation scripts don’t add value and error with #93 
* **Solution:** Drop ‘em!
* **Testing:** Ensure you can still commit to this repo.


packages/steelpenny/package.json
```diff
@@ -57,7 +57,6 @@
     "babel-core": "6.7.4",
     "babel-loader": "6.2.4",
     "babel-preset-es2015": "6.6.0",
-    "coins-validate": "2.0.6",
```
swashcap: `coins-validate` ensures there’s a `validate` npm script. We don’t need it anymore! (Also, pretty sure we dropped it for the dcrewrite feature in a99a7e3d4d0c46c9c57cbc4795b94c9f10ab8177.)

PR #98, opened 2017-01-05
Pennyfile update steelpenny file record tests

drewlandis:
#### asana
https://app.asana.com/0/8755179914444/238512866374556

# problem statement
The File Records route in Steel Penny is completely untested.

# solution
Add tests! At the same time fix some grammar/spelling errors and modernize some of the syntax.

# test
`npm test`


packages/steelpenny/src/test/integration/file-records.js
```diff
@@ -0,0 +1,102 @@
...
+      return apiClient.fileMeta.post(pennyData)
+      .then(rs => {
+        rs.data.data[0].should.deep.equal({ valid: true, metaUpload: true });
+      });
+    });
+
+    it('sends invalid hrrc_num in pennyData and returns a failing response', () => {
+      const pennyData = getNiftiPennyData();
+      pennyData.penny.validation.hrrc = '123-fakeHrrcNum';
+      return apiClient.fileMeta.post(pennyData)
```
jwtlake: _I think_ we might need to write some of these tests a little differently in mocha then what we've been doing with tape.  In tape we could use the `t.plan(count)` to ensure we got the number of responses we expected. (There may be something equivalent for mocha) 

I'm pretty sure some of these tests would fail to catch actual errors because we aren't testing both the `then` and `catch` responses to our api calls. 

For example if we send a bad hrc_numb and the server says it's valid then the response would come back as a `then` and we have no checks for that code path. The `it` function for this test would return without any failed assertions and it would incorrectly determine that this test had passed.

packages/penny-collector/test/helper.js
```diff
@@ -121,7 +121,7 @@ class helper {
 
   createPennyObj() {
     const dicom = {
-      templete: {
+      template: { // TODO: I think this should be template!
```
drewlandis: I meant to remove that comment before committing. All it was supposed to mean was that `templete` should be spelled `template`. Fixed. Sorry for the confusion!

packages/steelpenny/src/test/integration/file-records.js
```diff
@@ -0,0 +1,102 @@
...
+chai.should();
+
+describe('File Records', () => {
+  // Always wait for the server to be ready before beginning any tests
+  before('wait for server to be ready', () => (serverReady
+    .then(initApiClient)
+    .then(setApiClient)
+  ));
+
+  describe('send pennyData to upload route', () => {
```
drewlandis: Good suggestion, thanks! Fixed.

packages/steelpenny/src/test/integration/file-records.js
drewlandis: Good catch, thanks! Fixed.

packages/steelpenny/src/test/integration/file-records.js
```diff
@@ -60,13 +60,19 @@ describe('File Records', () => {
       return apiClient.fileMeta.post(pennyData)
       .then(rs => {
         rs.data.data[0].should.deep.equal({ valid: true, metaUpload: true });
+      })
+      .catch(() => {
```
jwtlake: I think we can exclude the catch in this case. We can let Mocha/chai catch the error and report back on it.

packages/steelpenny/src/test/integration/file-records.js
```diff
@@ -60,13 +60,19 @@ describe('File Records', () => {
...
+        should.fail(0, 1, 'Test failed when it should have passed');
       });
     });
 
     it('Should send invalid pennyData (bad hrrc_num) and return a failing response', () => {
       const pennyData = getNiftiPennyData();
       pennyData.penny.validation.hrrc = '123-fakeHrrcNum';
       return apiClient.fileMeta.post(pennyData)
+      .then(() => {
```
jwtlake: I talked to Cory about this cause I wasn't really sure myself. He recommended we try throwing an error in the `then` block. Mocha/chai will catch the error and report back that the test failed.

```javascript
it('Should send invalid pennyData (bad hrrc_num) and return a failing response', () => {
  const pennyData = getNiftiPennyData();
  pennyData.penny.validation.hrrc = '123-fakeHrrcNum';
  return apiClient.fileMeta.post(pennyData)
    .then(
      () => {
        throw new Error('Invalid pennyData request resolved');
      },
      (err) => {
        err.data.error.message.should.equal('invalid study hrrc');
      }
    );
});
```

packages/steelpenny/src/test/integration/file-records.js
drewlandis: Fixed. 01648f5

packages/steelpenny/src/test/integration/file-records.js
```diff
@@ -54,16 +54,16 @@ describe('File Records', () => {
     .then(setApiClient)
   ));
 
-  describe('send pennyData to upload route', () => {
-    it('sends valid pennyData and returns a successful response', () => {
+  describe('POST /fileMeta', () => {
+    it('Should send valid pennyData and return a successful response', () => {
```
drewlandis: Fixed. 9e8b7fd

packages/steelpenny/src/test/integration/file-records.js
drewlandis: Fixed. 3620f7c

packages/steelpenny/src/test/integration/file-records.js
drewlandis: Fixed. 9e8b7fd

jwtlake:
👍

PR #99, opened 2017-01-05
Fix DX Boom Error

drewlandis:
#### asana
https://app.asana.com/0/8755179914444/238512866374562

# Note:
Despite the branch name, this does NOT need to be a hotfix. It can just be merged in like normal and go through the regular release process. Also, this is pretty much a rubber stamper one line change. The diff looks huge because of the compiled code :(

# problem statement
When the dataexchange route hits an error, it is failing to print the error. The reason is that [boom.unauthorized](https://github.com/hapijs/boom#boomunauthorizedmessage-scheme-attributes) has a signature that looks like this:
* `Boom.unauthorized([message], [scheme], [attributes])`

where as _most_ other boom errors have a signature that look like this: 
* `Boom.badRequest([message], [data])`
* `Boom.forbidden([message], [data])`
* etc.

Long story short, I'm guessing that the original author expected the second argument to be `data`, but for `boom.unauthorized` it is not.

# solution
Update the `boom.unauthorized` call to directly print the error in the _first_ argument rather than pass it as the second argument.

# test
`npm test`


jwtlake:
Hmm interesting. Ya it seems like the `scheme` and `attributes` options are used to set the WWW-Authenticate header which I don't think we want. (I was trying to look into that header and I think it's for basic auth)

I wonder why the `data` option was omitted for this one. I'm curious to see what the rest of the group thinks about dumping the full error message here. I think boom sanitizes/omits the `data` field before replying to the client to prevent over exposure of server inner-workings.

PR #101, opened 2017-01-06
Pennyfile fix/remove some eslint-line-disable's

drewlandis:
#### asana
https://app.asana.com/0/8755179914444/239608707150224

# First!
Make sure that #98 is merged before this PR is merged! The reason being is that this PR was branched from the same branch as #98.

# problem statement
Atoms's eslint doesn't know where the root of a repo is and so cannot properly follow pattern matching rules that pertain to file paths. See issue here: https://github.com/AtomLinter/linter-eslint/issues/771.  The problem presents itself by throwing eslint errors in atom when we require devDependency modules in testing files even though this should be allowed by our air-bnb style guide.

# solution
As per discussion in the above mentioned issue, the solution is to add an `.eslintignore` dot file at the root of the repo. This let's Atom's eslint plugin know where the root is and perform pattern matching from this root. While I was fixing this, I also fixed/removed several other `eslint-disable-line` errors.

# test
`npm run tape`
Optional: Open the changed files in Atom and verify that no eslint errors are thrown.


jwtlake:
👍

PR #104, opened 2017-01-21
Major update to get PC and SP to play nicely with new schema

drewlandis:
#### asana
https://app.asana.com/0/8755179914444/25173395864330

## note:
Make sure that https://github.com/MRN-Code/db_schema/pull/250 is run in liquibase prior to testing!

# problem statement
We have updated our schema, now we need to update PC and SP to reflect those changes.

# solution
Update PC and SP to work with the new DB schema.

# test
in steelpenny: `npm test`
in penny-collector: `npm run tape`


swashcap:
> `npm run tape`

Run dat tape!

packages/steelpenny/src/lib/controllers/file-records.js
```diff
@@ -62,25 +62,27 @@ internals.validate = (bookshelf, pennyData) => {
...
   validate,
-}) => {
-  const FileRecord = Bookshelf.model('FileRecord');
-  const FileRecordParticipant = Bookshelf.model('FileRecordParticipant');
-
+}) => (
   // save record and detail with rollback protection
-  return Bookshelf.transaction(t => FileRecord.forge({
+  Bookshelf.transaction(t => Bookshelf.model('FileRecordGroup').forge({
+    hash: Math.random().toString(36).slice(2),
```
drewlandis: Nice catch, that was just a dummy data generator from when I started working on the PR. Fixed in fcc61f8

packages/steelpenny/src/lib/controllers/file-records.js
```diff
@@ -62,25 +62,27 @@ internals.validate = (bookshelf, pennyData) => {
...
+  }).save(null, { transacting: t })
+  .then(frgRs => Bookshelf.model('FileRecord').forge({
+    groupId: frgRs.get('id'),
     metadata: pennyData.penny,
-    deviceModalityDetailId: pennyData.penny.type.deviceModalityDetailId,
+    initialAcquisitionDate: new Date(),
+    modalityId: 1,
```
drewlandis: fixed: a11832c

packages/steelpenny/src/lib/controllers/file-records.js
```diff
@@ -70,37 +70,60 @@ internals.saveFileRecord = ({
...
+      studyId: validate.studyId,
+    }).save(null, { transacting: t }))
+
+    // rollback if dryrun
+    .tap(model => {
+      // if dry run, reject in order to trigger rollback
+      if (pennyData.dryRun) {
+        return Promise.reject({ valid: true, dryRun: true });
+      }
+      return Promise.resolve(model);
```
drewlandis: Since this line isn't part of the rollback, I might recommend adding a comment like: `// else, successfully save file record` Or something similar

packages/penny-collector/test/helper.js
```diff
@@ -125,7 +125,7 @@ class helper {
         files: [
           'SerieMR-0001-0001-1.dcm',
         ],
-        hash: utils.hash(Math.random().toString(36).slice(2)), // make tests idempotent
+        hash: utils.hash('Study20160504at081028rest_v01_R01'),
```
drewlandis: This may not be the exact hash signature that DICOM's end up using... This is okay as-is, but I might recommend instead using something like `hash: utils.hash('fakeDicomHash'),` just to make it more obvious that it is not legit

PR #105, opened 2017-01-23
Add dbmap and SSL info to the box

rssk:
# problem statement
- the 'box doesnt have some config, namely the dbmap and ssl info for coins

# solution
- `Put it in the box`


packages/coins-deposit-box/test.js
```diff
@@ -16,6 +16,13 @@ assert(isURL(depositBox.apiURL), 'API URL');
 assert(isString(depositBox.cookieName), 'Auth cookie name');
 assert(isString(depositBox.env), 'Environment');
 assert(isURL(depositBox.webURL), 'Web URL');
+assert(isString(depositBox.dbmap), '/coins/config/dbmap.json');
```
swashcap: `assert` [doesn’t compare strings](https://nodejs.org/api/assert.html#assert_assert_value_message). Use `assert.equal`.

Also, comparing equality is a brittle test. Why not ensure the filename has a `.json` extension and call it good?

swashcap:
Mind updating [steelpenny’s dbmap.json ref?](https://github.com/MRN-Code/quarterback/blob/master/packages/steelpenny/config/default.json#L19).

packages/coins-deposit-box/test.js
rssk: 👍

swashcap:
:+1: Looks great! Data Exchange tests aren't passing due to hard-coded values. Whatever.

Issue #106, opened 2017-01-25
DataExchange integration tests broken!


swashcap:
#### Problem

steelpenny’s DataExchange integration tests fail due to hard-coded values

#### Command

```shell
cd packages/steelpenny/
mocha --timeout=10000 test/integration/dataexchange.js
```

#### Output

```shell
  1) Dataexchange tests Group controller Group load filter statistics Load filter statistics without USER_INPUT as option label:
     TypeError: removeChangingProperty(...).should.deepEqual is not a function
      at doRequest.then (test/integration/dataexchange.js:424:51)
      at process._tickDomainCallback (internal/process/next_tick.js:129:7)

  2) Dataexchange tests Filter controller GET getFormAttributes with text option id:

      AssertionError: expected { Object (result) } to deeply equal { Object (result) }
      + expected - actual

                   "optionId": "30-39"
                   "optionLabel": "30-39"
                 }
                 {
      -            "count": 66
      +            "count": 67
                   "optionId": "40-49"
                   "optionLabel": "40-49"
                 }
                 {
                   "optionId": "60-69"
                   "optionLabel": "60-69"
                 }
                 {
      -            "count": 89
      +            "count": 88
                   "optionId": "70-80"
                   "optionLabel": "70-80"
                 }
                 {
                   "optionId": "DIABETIC AND HYPERTENSIVE"
                   "optionLabel": "DIABETIC AND HYPERTENSIVE"
                 }
                 {
      -            "count": 153
      +            "count": 155
                   "optionId": "ENROLLED"
                   "optionLabel": "ENROLLED"
                 }
                 {
               "_attributeLabel": "Gender"
               "_modalityId": 142
               "_options": [
                 {
      -            "count": 1273
      +            "count": 1275
                   "optionId": "Female"
                   "optionLabel": "Female"
                 }
                 {
      
      at Assertion.assertEqual (node_modules/chai/lib/chai/core/assertions.js:464:19)
      at Assertion.ctx.(anonymous function) [as equal] (node_modules/chai/lib/chai/utils/addMethod.js:40:25)
      at doRequest.then.result (test/integration/dataexchange.js:478:30)
      at process._tickDomainCallback (internal/process/next_tick.js:129:7)
```

#### Versions

* steelpenny: **5.1.1**
* OS: **OS X 10.12.2**
* Node: **7.4.0**
* npm: **4.1.1**

PR #107, opened 2017-01-26
Fix failing DataExchange tests

swashcap:
* **Problem:** See #106
* **Solution:** Strip out naughty properties from response and assertion objects
* **Testing:** `npm test` in _packages/steelpenny_


swashcap:
> So what's the long term solution?

Couple ways to approach it:

1. Set up a test database, as you suggested. We populate the appropriate tables with known data, then assert `/dataexchange` responses like we’ve done previously. This is the preferred method from what I understand.
2. Insert test data into the appropriate tables on test setup, then remove the test data on test teardown. I wouldn't want to do this with our production database, but it should be fine for dev/staging.
3. Assert the _shapes_ of the responses using something like [joi](https://www.npmjs.com/package/joi) or [json-schema](https://www.npmjs.com/package/json-schema). This should be less brittle because it doesn’t depend on precise values.

Want to go ahead and merge this?

jwtlake:
🎉

PR #108, opened 2017-01-26
Name matcher

rssk:
# problem statement
- Name-matcher was old, it needed a bunch of SSL config that it really shouldn't do.

# solution
- Rewrite sections, more of a 'remaster', and get rid of SSL and put it behind nginx.

# test
`npm test`

Compare the two commits, the initial commit is basically auto lint-fixes, and the second commit is the rewrite.


packages/name-matcher/utils/deleteNames.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
```
swashcap: _utils_ sounds like it’s part of the program/library, which this isn’t. Want to place it in _scripts_ or _bin_?

Also, might be nice to expose a npm script for this.

packages/name-matcher/server.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
```
swashcap: What would you think about moving _server.js_ → _lib/server.js_? This makes it a bit quicker to identify where source code is (all in one directory) and follows the pattern set by steelpenny and halfpenny.

packages/name-matcher/lib/nameMatcher.js
```diff
@@ -0,0 +1,193 @@
+'use strict';
+
+/**
+* nameMatcher object to expose methods for adding names and matching against them.
+* @module nameMatcher.
+*/
+
+const redis = require('redis');
+const Promise = require('bluebird');
```
swashcap: Don’t squash global `Promise`: import Bluebird as `bluebird`, or just the method you need:

```js
const { promisifyAll } = require('bluebird');
```

packages/name-matcher/lib/nameMatcher.js
```diff
@@ -0,0 +1,193 @@
...
+const database = {
+  first: '',
+  last: '',
+};
+
+/**
+* Function to retrieve the metaphone for a name.
+* Will retrieve the metaphone from the local collection of names if it exists,
+* or generate and store it if it does not.
+* @param {string} name - the name to generate a metaphone for
```
swashcap: Don’t need a `-` in the doc blocks as often seen in PHP land. JSDoc also shows non-primitives as capitalized, although it probably doesn't matter. I’d also suggest not describing the return value as `an array` when the type is clearly defined. Perhaps:

```diff
-* @param {string} name - the name to generate a metaphone for
+ * @param {string} name Target for metaphone generation
-* @param {object} localCollection - a collection of name-indexed metaphones (optional)
+ * @param {Object} [localCollection] Hash map of name-indexed metaphones
-* @return {array} an array of the generated metaphones, eg ["JKP", "AKP"].
+ * @returns {string[]} Generated metaphones
```

packages/name-matcher/lib/nameMatcher.js
```diff
@@ -0,0 +1,193 @@
...
+ * @return {[type]} [description]
+ */
+nameMatcher.clearNames = () =>
+Promise.all(
+  redisClient.delAsync(database.first),
+  redisClient.delAsync(database.last)
+).then(() => console.log('deleted first and last names')); // eslint-disable-line no-console
+
+
+/**
```
swashcap: What’s up with this floating docblock?

packages/name-matcher/lib/nameMatcher.js
```diff
@@ -0,0 +1,193 @@
...
+* or generate and store it if it does not.
+* @param {string} name - the name to generate a metaphone for
+* @param {object} localCollection - a collection of name-indexed metaphones (optional)
+* @return {array} an array of the generated metaphones, eg ["JKP", "AKP"].
+*/
+const getMetaphone = name => localStore[name.type][name.name]
+  || natural.DoubleMetaphone.process(name.name);
+
+
+/**
```
swashcap: Useless docblock. Care to clarify?

packages/name-matcher/lib/nameMatcher.js
```diff
@@ -0,0 +1,193 @@
...
+ * [clearNames description]
+ * @return {[type]} [description]
+ */
+nameMatcher.clearNames = () =>
+Promise.all(
+  redisClient.delAsync(database.first),
+  redisClient.delAsync(database.last)
+).then(() => console.log('deleted first and last names')); // eslint-disable-line no-console
```
swashcap: How is namematcher’s output being logged? Shouldn't we use hapi’s logging in the server?

packages/name-matcher/lib/nameMatcher.js
```diff
@@ -0,0 +1,193 @@
...
+  const match2 = mp1[1] === mp2[1];
+  // Match the inputs
+  if (match1 === true && match2 === true && jaro >= 0.6) {
+    // Both metaphones match exactly, and strings have distance of 0.6
+    return true;
+  } else if ((match1 || match2) && jaro >= 0.7) {
+    // One of the two metaphones match exactly, and strings have distance of 0.7
+    return true;
+  } else if (jaro >= 0.82) {
+      // The strings are highly similar
```
swashcap: Guessing `eslint --fix` didn’t get the comment spacing.

packages/name-matcher/lib/nameMatcher.js
```diff
@@ -0,0 +1,193 @@
...
+
+
+/**
+* parse all metaphones in a collection into array format
+* @param inputCollection {object} a collection of name-indexed metaphones
+* where the metaphones are colon-delimited (e.g. TYL:TYL)
+* @return {object} the localCollection with the delimited metaphones replaced with arrays
+*/
+const parseMetaphones = inputCollection =>
+_.reduce(inputCollection, (memo, metaphone, name) => {
```
swashcap: Prefer [`lodash.mapValues`](https://lodash.com/docs/#mapValues) when properties aren’t filtered/dropped.

packages/name-matcher/test/test.js
```diff
@@ -0,0 +1,77 @@
...
+    return nameMatcher.clearNames()
+    .then(() => {
+      const redisClient = redis.createClient();
+      return Promise.all(
+        redisClient.hgetallAsync('testfirstNames'),
+        redisClient.hgetallAsync('testlastNames')
+      ).then((names) => {
+        t.deepEqual(names, [null], 'Test DB cleared');
+      }).then(() => {
+        redisClient.quit();
```
swashcap: Looks like [`RedisClient#quit` uses a callback](https://github.com/NodeRedis/node_redis/blob/ff9b727609ea125919828f7373e40082fd432eec/examples/simple.js#L24)? Maybe chain in a `Promise.all` return?

packages/name-matcher/lib/nameMatcher.js
rssk: BB errors out on `promise.all` due to some incompatibility w/ native promises. Tried on 4.4.4, 6.9.1. 7.4.x and had the same issue, gave up :-(

packages/name-matcher/lib/nameMatcher.js
rssk: Yeah, this blocs are old, I'll correct them

packages/name-matcher/lib/nameMatcher.js
rssk: Missed these, will do

packages/name-matcher/lib/nameMatcher.js
rssk: Clear names is really only to be used in tests and from the cli, no route can call this, more of a utility

packages/name-matcher/lib/nameMatcher.js
rssk: ugh, yeah, and doesnt throw lint errors....

packages/name-matcher/lib/nameMatcher.js
rssk: 👍

packages/name-matcher/server.js
rssk: Agreed, more of a fan of stuffing it in lib

packages/name-matcher/utils/deleteNames.js
rssk: 👍

packages/name-matcher/test/test.js
rssk: Sure, I don't think it super matters, as the connection to the server is dropped regardless of the result from the target redis server, but it would look cleaner.

packages/name-matcher/lib/nameMatcher.js
swashcap: _Booo_ Bluebird.

packages/name-matcher/lib/nameMatcher.js
swashcap: Ah, cool.

packages/name-matcher/test/test.js
swashcap: Okay, fair enough. Their documentation on the method isn't super clear about the callback.

rssk: The documentation for node-redis is pretty poor, if you look at the quit code its a bit easier to see.

swashcap:
This looks great! Nice work. 🌈🥂🎊

drewlandis:
I'm a bit late to the game here. Noticed an issue with `npm start`. It is referencing the old location for `server.js`. Since this PR is closed, I opened an issue: https://github.com/MRN-Code/quarterback/issues/109

PR #110, opened 2017-01-28
Add namematcher executable

swashcap:
* **Problem:** namematcher can’t work as a service because there’s no executable
* **Solution:** Add _bin/namematcher.js_ and point to it in the _package.json_. (Bonus: address bad `npm start` script.)
* **Testing:**
  * Ensure `npm start` works
  * Ensure the executable works:
    ```shell
    cd packages/name-matcher
    npm link
    namematcher
    npm unlink # cleanup
    ```


drewlandis:
Looks good, thanks for the easy testing outlined above! 👍

PR #112, opened 2017-01-30
Fix DX Boom Error

drewlandis:
#### asana
https://app.asana.com/0/8755179914444/238512866374562

# Note:
Originally from this PR: https://github.com/MRN-Code/quarterback/pull/99. However, I closed #99 due to spiraling merge conflict errors having to do with compiled code diffs in half penny.

# problem statement
When the dataexchange route hits an error, it is failing to print the error. The reason is that [boom.unauthorized](https://github.com/hapijs/boom#boomunauthorizedmessage-scheme-attributes) has a signature that looks like this:
* `Boom.unauthorized([message], [scheme], [attributes])`

where as _most_ other boom errors have a signature that look like this: 
* `Boom.badRequest([message], [data])`
* `Boom.forbidden([message], [data])`
* etc.

Long story short, I'm guessing that the original author expected the second argument to be `data`, but for `boom.unauthorized` it is not.

# solution
Update the `boom.unauthorized` call to directly print the error in the _first_ argument rather than pass it as the second argument.

# test
`npm test`


drewlandis:
Comment from @jwtlake found in #99 
Hmm interesting. Ya it seems like the scheme and attributes options are used to set the WWW-Authenticate header which I don't think we want. (I was trying to look into that header and I think it's for basic auth)

I wonder why the data option was omitted for this one. I'm curious to see what the rest of the group thinks about dumping the full error message here. I think boom sanitizes/omits the data field before replying to the client to prevent over exposure of server inner-workings.

drewlandis:
@rssk What do you think about the above? I see 3 possible solutions, but I'm sure there are more...
1) Leave it as-is (e.g. print the error in the message).
```
throw boom.unauthorized(`Could not parse cookie. Error: ${JSON.stringify(error)}`);
```
2) Remove the error altogether.
```
throw boom.unauthorized('Could not parse cookie');
```
3) Convert `boom.unauthorized` --> `boom.badData` (or something else)
```
throw boom.badData('Could not parse cookie', error);
```

jwtlake:
Per Ross's feedback we went with option 2
`throw boom.unauthorized('Could not parse cookie');`
Rebased on master with fix

swashcap:
🎉

PR #113, opened 2017-01-30
Add login route option for cstac

rssk:
Cstac has no easy way to transfer login creds to the DB, put them in SP
hawk login, returned on user auth.

# problem statement ^

# solution ^

# test
`npm test`


swashcap:
:tada: It’s working!

PR #114, opened 2017-02-09
Penny/auth

jwtlake:
# Problem Statement
- PennyCollector is missing coins auth functionality
- PennyCollector routes on SteelPenny have auth disabled
- tusUploads bypass SteelPenny /files route
  - Connects to TUS server directly (via config tusUploadURL)

# Solutions
## SteelPenny
- Require Auth on `POST /file-records`
- Require Auth on `POST /files`
  - Route proxies to TUS server
  - All TUS uploads start with a POST and return with a tusUploadPoint url
  - Client then finishes the upload connecting directly to the TUS server via the tusUploadPoint
    - Initially tried intercepting all requests but ran into some problems with how HAPI handles HEAD only requests
  - Will need to add a nginx rule to block POST requests to TUS server from outside the network
    - This way uploads can only be started when first authorized through the /files route

## PennyCollector
- Created new HalfPenny class
   - Wraps HP lib
     - will work well with the new HP lib which has less predefined functions 
   - Contains all coins api logic
   - Standardize responses and errors
   - Provides helper functions to:
     - Login with client config (hardcoded mocatest user for now)
     - Maintain a current auth token
     - Re-establish a connection after network outages or expired auth tokens
     - Delay hp requests until connection is re-establish
- Updated PC Bootup
  - attempts to login on start
  - crash on fail (devops to handle retry?)
- Updated Validate and Archive to use new HP class
  - HP keeps track of connection status (connection issues cover network and auth problems)
  - If connection is lost HP suspends all future HP requests until it is able to reconnect
  - Process Flow
    - Check connection status
      - `connected = true` continue 
      - `connected = false` wait for connected event then continue
    - Make HP request 
    - If there is a connection error during the request
      - Update HP connection status
      - Reconnect then retry HP request
- Updates to config
  - Removed TUS upload url (part of HP class now)
  - Added `coinsArchive` flag to enable or disable coins archiving

## Testing
- HalfPenny tests should now cover all COINS api testing
- Archive and Validate tests shouldn't retest HP class
  - Override HP functions to test module flow and not HP
  - Avoid starting a SteelPenny server if not required for test
- Add a normal use integration test (EndToEnd test?)
- Clean up technical debt in tests
  - testHelper had a bunch of fs.sync helper functions -> Promises
  - Added code block descriptions
  - Using spawn instead of fork to start SP in tests to make debugging work better

## Linting Fixes
 - Fixed linting errors
 - Still haves some `no-restricted-syntax` for ForOfStatement
   - Fix in another PR
# Test
`npm test`
* Note: Make sure you have the [schema](https://github.com/MRN-Code/db_schema/pull/250) changes from the previous [PR](https://github.com/MRN-Code/quarterback/pull/104)


packages/penny-collector/src/errors.js
```diff
@@ -2,6 +2,22 @@
...
+    if (!(error instanceof Error)) {
+      throw new ReferenceError('Expected error to be an instance of Error');
+    }
+    super();
+    this.name = 'HalfPennyError';
+    this.type = type; // login, auth, config
+    this.error = error;
+    this.penny = penny;
+  }
+}
```
cdaringe: this is creative.

- does having an `errorInstance.error` property make sense?  consider that the javascript runtime looks for `.message` & `.stack` for all sorts of things to work well.
- should a `halfpenny` error belong here, vs in halfpenny?  i'm getting a strong sense of low cohesion where strong cohesion should be

packages/penny-collector/src/archive/coins.js
```diff
@@ -2,54 +2,33 @@
...
-});
-
-const archive = (penny, basePath, uploadURL) => {
+const archive = (penny, hp, basePath) => {
   // get files in penny object
   const filesToUpload = penny.files;
 
-  // start uploading each file
+  // start uploading each file in the penny
   return Promise.map(filesToUpload, (filename) => {
```
cdaringe: this is totally fine.  i will state, however, that sometimes it's kinda confusing assigning bluebird to Promise.  Promise is a builtin, and doing this squashes that global.  it also makes `.map` look like a naturally available method, which it isnt.  anyway, no change requested, but #opinions.

packages/penny-collector/src/archive/coins.js
```diff
@@ -2,54 +2,33 @@
...
     const fullFilePath = path.join(basePath, filename);
-    // #TODO: add zip step
-    return upload(penny, fullFilePath, uploadURL);
+    // check connection status and start upload
+    return hp.ensureConnection()
+    .then(() => hp.upload(penny, fullFilePath))
+    .catch((error) => {
+      // check for connection errors
+      if (hp.isConnectionError(error)) {
```
cdaringe: 👏 <== applause

woo hoo!  nice.  its worth mentioning that a single immediate retry is not very robust.  you may consider, for the sake of ease, not retrying at all, or using a more robust retry mechanism.  it will also reduce the boilerplate/redundancy here

packages/penny-collector/src/archive/coins.js
```diff
@@ -2,54 +2,33 @@
...
+        return hp.reconnect()
+        .then(() => hp.upload(penny, fullFilePath));
+      }
+
+      // rethrow other non-recoverable errors
+      throw error;
+    });
   },
-  { concurrency: 1 }); // upload one at a time
+  { concurrency: 1 }) // upload one at a time
```
cdaringe: 🔥 hot.

packages/penny-collector/src/archive/index.js
```diff
@@ -27,10 +35,11 @@ const archive = (penny, config) => {
   }
 
   // wait for all steps to complete
-  return Promise.all(archiveSteps).then((archiveResult) => {
+  return Promise.all(archiveSteps)
+  .then((archiveResult) => {
     // check for fails
     for (const step of archiveResult) {
-      if (step instanceof ArchiveError) {
+      if (step instanceof CustomError || step instanceof Error) {
```
cdaringe: if the base error extends `Error`, which it should if its an Error, really only one instanceof check is needed

packages/penny-collector/src/archive/index.js
```diff
@@ -27,10 +35,11 @@ const archive = (penny, config) => {
   }
 
   // wait for all steps to complete
-  return Promise.all(archiveSteps).then((archiveResult) => {
+  return Promise.all(archiveSteps)
```
cdaringe: shouldnt the archive steps throw or reject, in and of themselves?  it seems wonky to have to sniff if any of them errored, vs having control immediately go to a catch block, or traverse up the promise chain.

packages/penny-collector/src/validate.js
```diff
@@ -0,0 +1,36 @@
...
+// eslint-disable-next-line arrow-body-style
+const validate = (penny, hp) => {
+  // validate and upload metadata to server
+  return hp.ensureConnection()
+  .then(() => hp.validateAndMetaUpload(penny))
+  .then(({ valid, objectStoreIdentifier, metaUpload }) => {
+    const uploadResult = {
+      valid: valid || false,
+      objectStoreIdentifier: objectStoreIdentifier || false,
+      metaUpload: metaUpload || false,
```
cdaringe: what's metaupload?

packages/penny-collector/src/validate.js
```diff
@@ -0,0 +1,36 @@
...
+    };
+    return Object.assign({}, penny, uploadResult);
+  })
+  .catch((error) => {
+    // check for connection errors
+    if (hp.isConnectionError(error)) {
```
cdaringe: re: retry robustness

packages/penny-collector/test/HalfPenny/HalfPenny.spec.js
```diff
@@ -0,0 +1,154 @@
+'use strict';
+
+/* eslint-disable import/no-extraneous-dependencies */
+
+const test = require('tape');
```
cdaringe: ah, refreshing :)

im down w/ addding `spec` or `test` in filenames.  we didnt do that before, and sometimes it was a tiny bit confusing when utilities were adjacent!  👍

packages/penny-collector/test/archive/local.spec.js
```diff
@@ -68,46 +75,45 @@ test('archive local :: does not overwrite existing files', (t) => {
...
-
-      testHelper.emptyTestFolders();
+  })
+  // second archive
+  .then(() => testHelper.writeFilesFromPenny(penny, 'watch', 'second'))
+  .then(() => local.archive(penny, watchPath, localArchive))
+  // #TODO: need to figure out how to make the archiver fail out when file exists
+  // .then(() => {
+  //   t.fail('archive should not succeed when the file exists');
+  // })
```
cdaringe: any motion there?

packages/steelpenny/src/lib/app-routes/coins/files.js
```diff
@@ -1,22 +1,13 @@
...
     method: 'POST',
-    config: {
-      auth: false,
-      handler: {
-        proxy: {
-          host: 'localhost',
-          port: 7113,
-          protocol: 'http',
-          passThrough: true,
-          xforward: true,
```
cdaringe: this gave you trouble huh, the proxy bit?  the added complexity of initializing a file-request in one spot then redirecting to another for the upload external of the API seems rough.  i think its worth fighting for to make the proxy continue to work.  maybe pair programming with a teamster?  feed me the deets on what went down w/ that!

packages/steelpenny/src/lib/controllers/files.js
```diff
@@ -0,0 +1,65 @@
...
+    if (!tusMetadata.penny) {
+      return reply(boom.badData('expected penny in tus upload metadata', tusMetadata));
+    }
+
+    // parse penny
+    let penny;
+    try {
+      penny = JSON.parse(tusMetadata.penny);
+    } catch (exception) {
+      return reply(boom.badData('penny failed to parse to JSON', tusMetadata));
```
cdaringe: maybe, for clarity for the error consumer, just `malformed penny, tusMetadata`

packages/steelpenny/src/lib/utils/get-plugins.js
```diff
@@ -79,6 +79,7 @@ const plugins = [
   {
     register: require('./response-formatter.js'),
     options: {
+      excludePaths: ['/files'],
```
cdaringe: nice catch. i bet that took a while to find :)

cdaringe:
this looks like some big steps forward.  i have moderate concern for this seperate Halfpenny instance.  i think it's worth considering enhancing the existing HP vs wrapping it.  some stuff isnt browser friendly, i get that, but theres no reason the file upload additions couldn't use a `mixin`/`plugin` like scheme to enhance PC and keep cohesion up.  im thinking `halfpenny-collector-mixin` or something that receives an `HP` instance and adds goodies, **just as an idea**.  plus, lerna linking makes all of that quasi-easy (you know, when it works ;))

packages/penny-collector/src/HalfPenny.js
```diff
@@ -0,0 +1,255 @@
...
+const path = require('path');
+
+/**
+ * Halfpenny
+ * @class
+ *
+ * @param {Object} config
+ * @param {string} config.coinsApiUrl
+ */
+class Halfpenny extends EventEmitter {
```
jwtlake: Ahh ya this is a bit confusing in it's current state. We rewrote HalfPenny awhile back with a new minimal design [philosophy](https://github.com/MRN-Code/quarterback/tree/master/packages/halfpenny#philosophy).

Once we finish merging these PRs into `pennyfile` then I need to rebase with `master` so I have the latest version of HalfPenny. Which won't have things like
```javascript 
return this.api.fileMeta.post(pennyData)
```
instead i'll do
```javascript 
return this.hp.post('/filerecords',{})
```

I could rename the wrapper class but I think I'll hold off on changes for now as I kinda like your idea of a mix-in. I'm still trying to get my head around how that would be implemented tho. Would that code be defined in the HalfPenny project and then optionally required in along with the base version?

packages/penny-collector/src/archive/coins.js
jwtlake: It should [call](https://github.com/MRN-Code/quarterback/blob/fe2be74345ce9d0eb87190b8d59bdb7a77fa7485/packages/penny-collector/src/archive/coins.js#L19-L21) the `hp.reconnect` function before actually retrying. 

The reconnect [function](https://github.com/MRN-Code/quarterback/blob/fe2be74345ce9d0eb87190b8d59bdb7a77fa7485/packages/penny-collector/src/HalfPenny.js#L118-L140) tries to reconnect in defined increments via default [config](https://github.com/MRN-Code/quarterback/blob/fe2be74345ce9d0eb87190b8d59bdb7a77fa7485/packages/penny-collector/src/HalfPenny.js#L49).

packages/penny-collector/src/validate.js
jwtlake: The idea was to support a dry run of sorts. If you wanted to validate meta information but not actually store the meta information in the database. Maybe if we implement a UI at some point that needs realtime validation.

`metaUpload = true` means we created a new fileRecord
`metaUpload = false` means we didn't create one.

packages/penny-collector/src/validate.js
jwtlake: re: retry robustness :reply

packages/steelpenny/src/lib/controllers/files.js
jwtlake: Done. 6a5daf3b32c179f0303faba4aea6ca09c7ccf3ff

packages/penny-collector/src/archive/coins.js
cdaringe: sure, but this was a one off attempt.  it probably doesn't make sense to sniff for failed network issues on _every application code_ call, nor, just an arbitrary subset.  rather, you'd probably want a defacto behavior built into some network calls, which would be handy to enable w/ some sort of retry flag that would work w/ any given XHR or fetch.  library code.  not all calls need it, just food for thought.  to me, i guess, it somewhat felt like the `.isConnectionError` belongs inside the `.upload` method.

follow your heart, especially if it works fine now.

packages/penny-collector/src/validate.js
cdaringe: oh right.  meta!  at first i thought, is that sufficiently descriptive? then i supposed that this is in context of a penny.

packages/steelpenny/src/lib/controllers/files.js
```diff
@@ -41,7 +41,7 @@ module.exports.post = {
     try {
       penny = JSON.parse(tusMetadata.penny);
     } catch (exception) {
-      return reply(boom.badData('penny failed to parse to JSON', tusMetadata));
+      return reply(boom.badData('malformed penny', tusMetadata));
```
cdaringe: #dope

packages/penny-collector/src/errors.js
jwtlake: Haha "creative" I know what that means.. ;) 

Thanks for the feedback tho. Seems like this might be a bad idea.

So what I was trying to accomplish here.. 
- I was trying to isolate all the error handling logic to the module that they related to. So all error checking logic for validate actions would be done in that module.
- The custom error classes were an attempt to easily convey what went wrong in a standardized way to the [main promise chain](https://github.com/MRN-Code/quarterback/blob/penny/auth/packages/penny-collector/src/PennyCollector.js#L27-L47) and later in the error logging. (It also made [tests](https://github.com/MRN-Code/quarterback/blob/penny/auth/packages/penny-collector/test/validate/validate.spec.js#L128-L131) pretty easy)

I was trying to avoid 
```javascript
handleEvent(filePath) {
    this.logger.info(`processing new file: ${filePath}`);
    models.createPenny(filePath)
      .then(penny => validate(penny, this.hp))
      .catch(error => {
         // what kind of error is this? what should we do next?
       })
      .then(penny => archive(penny, this.hp, this.config.get()))
      .catch(error => {
         // what kind of error is this? what should we do next?
       })
      .then(penny => trash(penny, this.config.get()))
      .catch(error => {
         // what kind of error is this? what should we do next?
       })
  }
```

I guess its really not that complicated atm but I was trying to build myself some flexibility for down the road. Are we going to do another meetup this week? It would be nice to bounce some design ideas off you.

PR #115, opened 2017-02-21
Enhancement/improve error logging

jwtlake:
# problem statement
- Steelpenny errors are very difficult to read in console. 

# research
The reason errors are difficult to read is because we are stringifying errors in our [global-error-handler.js](https://github.com/MRN-Code/quarterback/blob/master/packages/steelpenny/lib/utils/global-error-handler.js#L8-L14)  so that they can be consumed by [good](https://www.npmjs.com/package/good) . The output can then be routed to both the [console](https://github.com/MRN-Code/quarterback/blob/master/packages/steelpenny/lib/utils/get-good-config.js#L9-L10) and to a [logfile](https://github.com/MRN-Code/quarterback/blob/master/packages/steelpenny/lib/utils/get-good-config.js#L16-L18).

I was able to find a few post on people trying to pass `Error` objects through `good` but from the responses it doesn't seem like it's something the module creators thought was a very good idea.
https://github.com/hapijs/hapi/issues/2744
https://github.com/hapijs/good/issues/414

Overall what we have works for logging to file, the big issue seems to be readability while debugging. So we kinda have two options that I can think of. 

1. Format the string output of good to have the spacing we want (in just the console reporter)
   * Use [good-reporter](https://www.npmjs.com/package/good-console) as a templet and make a custom console reporter that formats our error objects as desired
   * The [SafeJson](https://github.com/hapijs/good-squeeze#safejsonoptions-stringify) module allows you to append a separator. We might be able to use a character that escapes when in console somehow? Default is already the character I would have tried.
2. Dump the error directly to console like the good lord intended.

# solution
- Use console.error to output `Error` objects directly to console so built in formating kicks in.
- Updated good versions and updated to newer syntax which allows for chaining.

# test
- `npm test`
- break some stuff and make sure logging to the console and to files looks good.


packages/steelpenny/lib/controllers/auth/keys.js
```diff
@@ -185,7 +185,7 @@ module.exports.delete = {
         .state(casCookieUtils.cookieName, casCookieUtils.invalidate())
       )
       .catch(error => {
-        request.log(['error', 'logout'], error);
+        request.log(['error', 'logout'], error.message);
```
swashcap: Why only log the error’s message here? Can’t hap’s logger handle the entire `Error`?

packages/steelpenny/lib/utils/global-error-handler.js
```diff
@@ -1,17 +1,21 @@
...
+/* eslint-disable no-console */
+
 // uncaughtExceptions and unhandledRejections are re-routed here s.t. the server logger
 // can record the incident before crashing
 module.exports.register = (server, options, next) => {
   server.expose({
     handleUncaught: (err) => {
       server.log(
-        ['error'],
+        ['global-error'],
```
swashcap: What’s the significance of `global-error` versus `error` here? Does `global-error` need to be added in every log call’s tags?

packages/steelpenny/lib/utils/global-error-handler.js
```diff
@@ -1,17 +1,21 @@
...
         {
           message: err ? err.toString() : '(empty Error)',
           stack: err.stack,
         }
       );
+      console.error(err.message);
```
swashcap: Just `console.error(err)` here: `console.error` logs the message and stack trace. Double logging leads to confusing repetition:

```real
> var e = new Error('wat')
> console.error(e)
Error: wat
    at repl:1:15
    at ContextifyScript.Script.runInThisContext (vm.js:23:33)
    at REPLServer.defaultEval (repl.js:340:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.onLine (repl.js:537:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:189:7)
    at REPLServer.Interface._onLine (readline.js:238:10)
    at REPLServer.Interface._line (readline.js:582:8)
undefined
> console.error(e.message); console.error(e.stack);
wat
Error: wat
    at repl:1:9
    at ContextifyScript.Script.runInThisContext (vm.js:23:33)
    at REPLServer.defaultEval (repl.js:340:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.onLine (repl.js:537:10)
    at emitOne (events.js:101:20)
    at REPLServer.emit (events.js:189:7)
    at REPLServer.Interface._onLine (readline.js:238:10)
    at REPLServer.Interface._line (readline.js:582:8)
undefined
```

packages/steelpenny/lib/utils/register-uncaught-exceptions.js
```diff
@@ -10,9 +12,9 @@ module.exports.register = (server, options, next) => {
       try {
         handler.handleUncaught.call(handler, err);
       } catch (suberr) {
-        console.error(err.message); // eslint-disable-line
-        console.error(err.stack); // eslint-disable-line
-        console.error([  // eslint-disable-line
+        console.error(err.message);
+        console.error(err.stack);
+        console.error([
           'the server error logger was unable to log the uncaught error.',
```
swashcap: Please put strings in a single line for easier string searches. For now you’ll need to add an ESLint disable comment, but future versions of our style guide ignore max-line-length for strings.

You should be able to rewrite this:

```diff
- console.error(err.message);
- console.error(err.stack);
- console.error([
-   'the server error logger was unable to log the uncaught error.',
-   'the logger experienced the following error:',
-   `${suberr.message}`,
+ /* eslint-disable max-len */
+ console.error(
+   'the server error logger was unable to log the uncaught error. the logger experienced the following error:',
+   suberr
  );
+ /* eslint-enable max-len */

File: packages/steelpenny/lib/utils/register-unhandled-rejections.js [hunk 5]
```diff
@@ -12,9 +15,9 @@ module.exports.register = (server, options, next) => {
       try {
         handler.handleUncaught.call(handler, err);
       } catch (suberr) {
-        console.error(err.message); // eslint-disable-line
-        console.error(err.stack); // eslint-disable-line
-        console.error([  // eslint-disable-line
+        console.error(err.message);
```
swashcap: Same thing as _lib/utils/register-uncaught-exceptions.js_: single `console.error` call. Also, this file looks pretty similar to the unhandled rejection listener. DRY it out?

packages/steelpenny/lib/controllers/auth/keys.js
jwtlake: I removed this commit after implementing `good-error`. The issue wasn't `hapi` logging it was some of the `good` modules.

packages/steelpenny/lib/utils/global-error-handler.js
jwtlake: I removed this commit after implementing good-error. The tag `global-error` was just something I made up so I could tell the `good` console reporter to ignore server.logs from this code path because we were directly console.error()ing here. Otherwise we would get two error outputs to console.

packages/steelpenny/lib/utils/global-error-handler.js
jwtlake: Sounds good.

packages/steelpenny/lib/utils/register-uncaught-exceptions.js
jwtlake: To get to this code path we would have to have a `uncaught-exception` that throws an additional error when try to pass the error off to `global-error-handler.js`. That's why we have two error outputs. 

`suberr` is the original error
`error` is the error from `global-error-handler.js` 

We aren't doing much inside `global-error-handler.js` so I can't really think of a situation that would actually cause this code path to be reached. I implemented the suggested changes with some slight tweeks to keep the original functionality. (Let me know if you think we can cut the extra error handling out)

```
console.error(err);
/* eslint-disable max-len */
console.error(
  `the server error logger was unable to log the uncaught error. the logger experienced the following error:
  ${suberr}`
);
/* eslint-enable max-len */
```

I used a string template to get an extra line brake before printing the error.

jwtlake:
So I re-read some stuff Friday while I was attempting to answer some of your questions and I think I might have a better solution. Changes have been rebased into this branch.

With this new approach we should be able to hapi.log (server,response, etc...) both custom messages and error objects. Then downstream we have three `good` reporters. One console reporter and two file reporters.

### Console Reporter Flow

1. `good-squeeze` filter stream utility
2. `good-console` updated version has error object formating built in. (It's not quite as pretty as console.error output see below)

### File Reporters Flow

1. `good-squeeze` filter stream utility
2. `good-errors` stream utility that adds error object handling
3. `good-file` doesn't support error objects on its own and would output `{}` without `good-errors`


If we go this route our logging outputs will look like this
```
// (console output)
170303/223058.134, [log,error] message: getaddrinfo ENOTFOUND devdbcoin.mind.unm.edu devdbcoin.mind.unm.edu:5432, stack: Error: getaddrinfo ENOTFOUND devdbcoin.mind.unm.edu devdbcoin.mind.unm.edu:5432
  at errnoException (dns.js:28:10)
  at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:76:26)

// (file output)
{"event":"log","timestamp":1488580033707,"tags":["error"],"data":{"code":"ENOTFOUND","errno":"ENOTFOUND","syscall":"getaddrinfo","hostname":"devdbcoin.mind.unm.edu","host":"devdbcoin.mind.unm.edu","port":5432,"stringifyable":{"name":"Error","stack":"Error: getaddrinfo ENO    TFOUND devdbcoin.mind.unm.edu devdbcoin.mind.unm.edu:5432\n    at errnoException (dns.js:28:10)\n    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:76:26)","message":"getaddrinfo ENOTFOUND devdbcoin.mind.unm.edu devdbcoin.mind.unm.edu:5432"}},"pid":83320}

```

Alternatively if we stick with console.error directly to console it would look like this

```
// console.error()
{ Error: getaddrinfo ENOTFOUND devdbcoin.mind.unm.edu devdbcoin.mind.unm.edu:5432
    at errnoException (dns.js:28:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:76:26)
  code: 'ENOTFOUND',
  errno: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'devdbcoin.mind.unm.edu',
  host: 'devdbcoin.mind.unm.edu',
  port: 5432 }
```

The main reason we have all this weirdness is because the older versions of `good` didn't support error objects natively so to get around this we were [pulling error attributes out and logging as string](https://github.com/MRN-Code/quarterback/blob/master/packages/steelpenny/lib/utils/global-error-handler.js#L8-L14). This worked but it made the `good-console` output very difficult to read. With the newer versions of `good` it seems like we don't have that limitation so we can just [log.error objects](https://github.com/MRN-Code/quarterback/blob/a568bdf0dfef4ba12e44111f5a9cb3c789f82f77/packages/steelpenny/lib/utils/global-error-handler.js#L9-L12) if needed. The one downside is that `good-console` doesn't format the error object ouput quite as well as a direct console.error. StackTrace has line breaks but the rest of the error attributes are output as one line.

We could still add an additional console.error like I had originally.
```
server.log(['error'], err);
console.error(error);
```
but then we would have double output to console. (one from `good` and one from console.error). To get around this we could add a new error tag that we could use as a filter in our `good` config so the good console reporter knows to ignore log calls that we have already console.error'ed the output for.
```
server.log(['global-handled-error'], err);
console.error(error);
```

packages/steelpenny/lib/utils/global-error-handler.js
swashcap: :+1:

swashcap:
> The main reason we have all this weirdness is because the older versions of good didn't support error objects natively

Ah, very interesting.

I’d prefer to limit our logging configuration as much as possible and let _good_ do its thing. This should hopefully help us avoid situations like this in the future.

> We could still add an additional console.error like I had originally.

I think the overhead of double console output and two different error tags will confuse future developers. How about we go with the default _good-console_ output? If its output ends up being troublesome we can approach the output via a custom log parser or something.

swashcap:
:+1: Nice work!

PR #118, opened 2017-03-01
Penny/file handler

jwtlake:
place holder for notes

# problem statement
- What is the problem or feature request.

# solution
- How does this pull request fulfill or solve the problem.

# test
- How can the code reviewer test this change.


packages/coins-archiver/lib/routes.js
```diff
@@ -0,0 +1,50 @@
...
+      // Archive file
+      .then(({ penny, modalityConfig }) => CoinsArchiver.archiveTusFile({
+        tempFileName: request.payload.tempFileName,
+        filename: request.payload.filename,
+        penny,
+        modalityConfig,
+      }))
+
+      // Reply with results
+      .then(newLocation => reply(newLocation).code(201))
```
cdaringe: Should we be telling clients this info?

packages/coins-archiver/lib/routes.js
```diff
@@ -0,0 +1,50 @@
...
+      .catch(error => reply(Boom.badImplementation(error)));
```
cdaringe: If anything above fails, is there any cleanup or roll back we need to consider?  This is looking pretty good for the most part!

packages/coins-archiver/test/integration/main.js
```diff
@@ -0,0 +1,77 @@
...
+    .then(() => config.writeConfigPaths(testConfig))
+    .then(() => server.start())
+    .should.be.fulfilled;
+  });
+
+  let tusUploadTestFileDetails;
+
+  it('moves TUS files to archive correctly', () => {
+    const pennyObject = penny.createFromTemplete.nifti.valid();
+    const expectedpath = `${testConfig.archivePath}/human/nifti/unknown-scanner/JPHILLIPS/TECHDEV_20-0052/M80327555/20160809at115703/rest_v01_R01/JPHILLIPS+TECHDEV_20-0052+M80327555+20160809at115703+rest_v01_R01+CM.nii`;
```
cdaringe: Is this a real data? :X. Even the ursi study combo here may warrant a rebase.  Not sure.  Check with the mk or bm

packages/coins-archiver/test/integration/main.js
```diff
@@ -0,0 +1,77 @@
...
+
+    // simulate a recently finished tus upload
+    return helper.createTusUploadFile(testConfig.tusUploadPath)
+    .then((details) => {
+      tusUploadTestFileDetails = details; // save file info for testing later
+      return details;
```
cdaringe: _details_.

packages/coins-archiver/test/unit/CoinsArchiver.js
```diff
@@ -0,0 +1,102 @@
...
+const config = require('../utils/config');
+const penny = require('../utils/penny');
+const modality = require('../utils/modality');
+const CoinsArchiver = require('../../lib/plugins/CoinsArchiver/archiver');
+
+chai.use(chaiAsPromised);
+chai.should();
+
+describe('CoinsArchiver', () => {
+  before('create test paths', () => {});
```
cdaringe: I assume this is cool, or is singing meant to be here

cdaringe: Something

packages/steelpenny/src/tus/tus-server.js
```diff
@@ -1,12 +1,40 @@
...
-const host = '0.0.0.0'; // '127.0.0.1';
-const port = 7113;
+// init webrequest module
+const Axios = axios.create({
+  baseURL: 'http://127.0.0.1:3210/', // #TODO: add config
+  timeout: 1000,
+});
+
+// register event hook
+server.on(tus.EVENTS.EVENT_UPLOAD_COMPLETE, (event) => {
```
cdaringe: `UPLOAD_COMPLETE`? why EVENT prefix?

packages/steelpenny/src/lib/models/Modality.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+module.exports = (bookshelf) => bookshelf.extend({
+  tableName: 'mrs_modalities',
+  idAttribute: 'modality_id',
+  fileRecordModality: function() {
```
cdaringe: should a modality be tightly coupled to a file record?

packages/steelpenny/src/tus/tus-server.js
```diff
@@ -1,12 +1,40 @@
...
-    path: '/files'
+  path: '/files',
+  // Note: this appears to sets the http route and
+  // the local fs path gets created in steelpenny root
 });
 
-const host = '0.0.0.0'; // '127.0.0.1';
-const port = 7113;
+// init webrequest module
+const Axios = axios.create({
```
cdaringe: would you consider something like `tusXHRClient`?  This reads like theres a new constructor in scope

packages/steelpenny/src/tus/tus-server.js
```diff
@@ -1,12 +1,40 @@
...
+  const tempFileName = event.file.id;
+  const metaData = decode(event.file.upload_metadata);
+
+  // merge event data to create payload object
+  const payload = Object.assign({}, { tempFileName }, metaData);
+
+  // notify fileHandler that a new file is ready for processing
+  Axios.post('/file', payload)
+  .catch((error) => {
+    console.log(error);
```
cdaringe: If we just log, do we run the risk of having a bunch of ready to process things that never get processed?  Do we need to gracefully retry somehow?

cdaringe:
Nice work! 🎉

packages/coins-archiver/lib/index.js
```diff
@@ -0,0 +1,64 @@
...
+        client: 'pg',
+        connection: {
+          host: config.get('dbServer.host'),
+          database: config.get('dbServer.database'),
+          user: config.get('dbServer.user'),
+          password: config.get('dbServer.password'),
+          charset: 'utf8',
+        },
+      },
+      plugins: ['registry'],
```
jwtlake: You know.. honestly I'm not sure.. 

`hapi-bookshelf-models` has it as a required [option](https://github.com/lob/hapi-bookshelf-models#options) but I don't see anything that explains why. I just mimiced what I saw in [steelpenny](https://github.com/MRN-Code/quarterback/blob/master/packages/steelpenny/lib/utils/get-plugins.js#L56).

packages/coins-archiver/lib/index.js
```diff
@@ -0,0 +1,64 @@
...
+      base: require('../../steelpenny/src/lib/utils/get-base-model.js'),
```
jwtlake: Ah ya so this was totally a hack to get things rolling.

We've been kicking around the idea of refactoring the bookshelf models out of `SteelPenny` and into their own project so other projects can require in models without a bunch of overhead or code duplication.

Thoughts?

packages/coins-archiver/lib/index.js
```diff
@@ -0,0 +1,64 @@
...
+/* eslint-enable global-require */
+
+server.register(plugins.concat(routes), (registerError) => {
+  if (registerError) {
+    throw registerError;
+  }
+
+  server.start((startError) => {
+    if (startError) {
+      throw startError;
```
jwtlake: Fixed!
2222557aad87b73196e9746cf1de9ce0fb77c6a6

packages/coins-archiver/lib/index.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+    console.log(['startup'], `server running at: ${server.info.uri}`);
```
jwtlake: Whoops! Sloppy copy job here. Removed. f65a2ae1c5d3a12dbe8d11622036ece567ff071d

packages/coins-archiver/lib/index.js
cdaringe: huh. ok! 👍

packages/coins-archiver/lib/index.js
cdaringe: yea, that makes good sense.  it's probably not much work either! maybe a days worth.

packages/coins-archiver/lib/routes.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const Joi = require('joi');
+const Boom = require('boom');
+
```
jwtlake: Fixed! 9bf62b554ba36e967c2a8d81b8a0a8d99bd78d09

packages/steelpenny/src/tus/tus-server.js
jwtlake: It just seems to be how the library is structured. Their [events example ](https://github.com/tus/tus-node-server#events) 
```
const Server = require('tus-node-server').Server;
const EVENTS = require('tus-node-server').EVENTS;
```
I just figured I'd skip the extra declaration because there was only one event hook that made sense using. I'm not super opinionated about it. Any recommendations?

cdaringe: nope. sorry, this is fine. for whatever reason, i thought we defined that

packages/steelpenny/src/lib/models/Modality.js
jwtlake: This is probably a poor naming choice on my part.`fileRecordModality` is a intersection table that stores archive config for a modality.

FileRecord -> Modality -> FileRecordModality

So a FileRecord has one Modality
Each Modality has one FileRecordModality record

Does that answer your question or did I miss the point?

packages/steelpenny/src/tus/tus-server.js
jwtlake: Ya thats a good point. There are a lot of handoffs with our multi service approach which makes things a bit tricky. 

As it sits now if something failed here I would loose all the meta information that was passed in the TUS headers (IE: objectSourceIdentifier and OriginalFileName) which would make it pretty difficult to manually fix.

Hmm..

packages/steelpenny/src/tus/tus-server.js
jwtlake: Ya makes sense. 532640bfb5d0e52c26bd15e52d3541d3f5cc794d

packages/coins-archiver/test/unit/CoinsArchiver.js
jwtlake: Same tests. Less singing. 407dce9b8958bd1b00eba125d478d70ed1e53865

packages/coins-archiver/test/integration/main.js
jwtlake: Good catch.. :( 
This has bought up a hole can of worms as we have been using this for awhile in PC.

packages/coins-archiver/lib/routes.js
```diff
@@ -0,0 +1,50 @@
...
+    },
+    handler: (request, reply) => {
+      const FileRecord = request.server.plugins.bookshelf.model('FileRecord');
+      const CoinsArchiver = request.server.plugins.CoinsArchiver;
+
+      // Get FileRecord and Modality Config
+      return FileRecord.where('object_store_identifier', request.payload.objectStoreIdentifier)
+      .fetch({ withRelated: ['fileRecordModality'], require: true })
+      .then(fileRecord => ({
+        penny: fileRecord.get('metadata'),
```
jwtlake: ¯\_(ツ)_/¯ Different naming conventions between different projects. PennyCollector calls the metaData object `penny` but our database uses the column name `metadata`.

Do you think its worth standardizing across all the projects? If so which would would you go with?

packages/coins-archiver/lib/routes.js
jwtlake: The client should only be the TUS server. The plan was to use nginx to block everything else. However TUS doesn't need this information so I went ahead and removed it. c6fda77619d5549e0050c59df7976e2fd64ce710

packages/coins-archiver/lib/routes.js
cdaringe: oh i c.  ya, totally unify!  i think it will pay off in the long run!

i'd go with `penny`.  `penny` has much more meaning than `metadata`.  `metadata` == `details` == `data` == `everything` == `ambiguous`.  how do you feel about it?

packages/coins-archiver/lib/routes.js
```diff
@@ -0,0 +1,50 @@
...
+const Boom = require('boom');
+
+exports.register = (server, options, next) => {
+  server.route({
+    method: 'POST',
+    path: '/file',
+    config: {
+      validate: {
+        payload: {
+          tempFileName: Joi.string().required(),
```
jwtlake: > Letting the client set this feels like a security or integrity vuln.

So the client is the TUS server. We planned to use nginx to block everything else.  

> Can the server alone own this?

I can't think of anyway to get around this with the current multi-project design. TUS has to tell coins-archiver the name of the temporary file it created and what meta information was passed along with the upload.

packages/coins-archiver/lib/routes.js
```diff
@@ -0,0 +1,50 @@
...
+          filename: Joi.string().required(),
+          objectStoreIdentifier: Joi.string().required(),
```
jwtlake: `objectStoreIdentifier` is an uuid for the `file-record` table. 

We create the uuid [here](https://github.com/MRN-Code/quarterback/blob/penny/auth/packages/steelpenny/src/lib/controllers/file-records.js#L83) which is then [saved](https://github.com/MRN-Code/quarterback/blob/penny/auth/packages/steelpenny/src/lib/controllers/file-records.js#L106) to the DB and [returned](https://github.com/MRN-Code/quarterback/blob/penny/auth/packages/steelpenny/src/lib/controllers/file-records.js#L127) to PennyCollector to use when uploading the actual file. 

It's kinda being used as an extra security token between meta validation and file uploads.

PR #121, opened 2017-03-07
Fix login response shape bug

swashcap:
#### asana

https://app.asana.com/0/inbox/37131652076105/288455728367427/288465675195049

# problem statement

`POST /auth/keys` returns a string-ified `user` property

# solution

Un-string-ify the `user` property. Add a test!

# test

Jump on dat VPN. Then:

```shell
git fetch --all
git checkout bug/auth-keys-user
mocha --timeout=10000 test/integration/keys.js
```


packages/steelpenny/test/integration/keys.js
```diff
@@ -73,6 +76,21 @@ describe('POST keys (login)', () => {
       })
     );
 
+    it('Should respond with user', () => loginPromise.then(
+      ({ data: { data: [response] } }) => {
+        response.should.to.have.property('user');
```
jwtlake: Is the extra `to` necessary? It looks like we are mashing up the `expect` and `should` [syntax](http://chaijs.com/guide/styles/).

```
expect(foo).to.be.a('string');
foo.should.be.a('string');
```

Looks like we are doing `should.to` in other tests from this file tho so its consistent at least.  ¯\_(ツ)_/¯

swashcap: Yeah, just copied it. chai's documentation doesn't clearly explain the chaining syntax, or we're using an older syntax, or…something? I'd love to move to `chai#assert`-style at some point.

swashcap:
> Looks good to me.

@jwtlake Cool, you want to merge the branch and publish?

PR #123, opened 2017-03-16
Maildaemon

rssk:
# problem statement
- The maildaemon is in legacy php... :-(

# solution
- rewrite it to use notifications via a PG notification channel. There is also a db_schema for this.

# test
- `npm test`


packages/coins-mail-daemon/lib/mailProcessor.js
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Grabs the first 1000 pieces of mail to send that have a send time
+ * after now, a promise is then returned wether the rows send correctly or not
+ * @param  {object} bookshelf   bookshelf instance
+ * @param  {object} transporter nodemailer transporter
+ * @return {Promise}            promise that resolves on all rows mail being proccessed
+ */
+const processMail = (bookshelf, transporter) =>
+  bookshelf.knex('mrs_mail').where('send_time', '<', moment().format()).whereNull('sent', null).limit(1000)
```
rssk: While a good idea, I think its a bit overkill for just that param

packages/coins-mail-daemon/lib/mailProcessor.js
```diff
@@ -0,0 +1,75 @@
...
+  .then(rows => Promise.all(rows.map(row => processRow(row, bookshelf, transporter).reflect())));
```
rssk: Old stuff, not needed, dorpin

packages/coins-mail-daemon/lib/mailProcessor.js
```diff
@@ -0,0 +1,75 @@
...
+
+module.exports = {
+  start(bookshelf, transporter) {
+    processMail(bookshelf, transporter) // go through mail on startup
```
rssk: The query _should_ grab from descending IDs so no piece of mail should be missed, but! there is a chance that there'd be a piece of mail that'd be missed by the 1k limit in the 1001 case +1k, +1k, etc... I don't think this is a concern for our current load, but I'll also explicitly sort the query

packages/coins-mail-daemon/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "publish-major": "npm run preversion && npm version major && git push origin master --tags && npm publish",
+    "start": "node server/index.js",
+    "test": "nyc --reporter=lcov node test/index.js",
+    "test:coverage": "istanbul cover test/index.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/MRN-Code/quarterback.git"
+  },
+  "author": "rkelly@mrn.org",
```
rssk: probably makes more sense

packages/coins-mail-daemon/test/index.js
```diff
@@ -0,0 +1,252 @@
...
+const tracker = mockKnex.getTracker();
+tracker.install();
+box.dbmap = '../test/stubs/dbmap.json';
+
+let mailStub;
+let dbStub;
+
+const mailSendTime = moment().format();
+
+// async, so not always called in order
```
rssk: 👍

packages/coins-mail-daemon/test/index.js
```diff
@@ -0,0 +1,252 @@
...
+};
+
+tape('processes and sends mail', (t) => {
+  // the plan count is confusing:
+  // Both the sendMail and on('query') fns get called mult times
+  // the count reflects this
+  t.plan(14);
+
+  const transStub = (config) => {
+    if (_.isEqual(config, configMock)) {
```
rssk: It checks the nodemailer config shape, I'll add a `t` for it

PR #126, opened 2017-03-17
Remove dateAdded schema restriction

mstone121:
This column was added along with a few others in https://github.com/MRN-Code/quarterback/pull/122. However this column has not yet been merged when the others have.


jwtlake:
User tests are now passing on a fresh coins db templet. Looks good to me!

